PR #1, opened 2026-01-04
Implemented OpenAI compatible API configuration.

DeclineThyself:
image

This will default to OpenRouter, and migrate settings.
```typescript
// Load API settings
const apiURL = await database.getSetting('openai_api_url') ?? OPENROUTER_API_URL; //Default to OpenRouter.
const apiKey = await database.getSetting('openai_api_key') ?? await database.getSetting('openrouter_api_key'); //Migrate API key location.
```
I didn't rename `src/lib/services/ai/openrouter.ts` so this commit is easier to review.

Should I move `debounce` to a new `utils.ts`, or `import lodash`?
```typescript
// Import lodash?
// import { debounce } from 'lodash';
// If not, Where should this go?
let timeout: number|null = 0;
const debounce = (to_run: Function, time: number = 100) => {
  timeout && clearTimeout(timeout);
  timeout = setTimeout(() => {
    timeout = null;
    to_run();
  }, time);
};
```

Would you like any changes made to this PR?

### DIFF
--- a/src/lib/components/settings/SettingsModal.svelte
+++ b/src/lib/components/settings/SettingsModal.svelte
@@ -2,7 +2,7 @@
   import { onMount } from 'svelte';
   import { ui } from '$lib/stores/ui.svelte';
   import { settings, DEFAULT_SERVICE_PROMPTS } from '$lib/stores/settings.svelte';
-  import { OpenRouterProvider } from '$lib/services/ai/openrouter';
+  import { OpenAIProvider, OPENROUTER_API_URL } from '$lib/services/ai/openrouter';
   import type { ModelInfo } from '$lib/services/ai/types';
   import type { ThemeId } from '$lib/types';
   import {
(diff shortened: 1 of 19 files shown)


surfacecoaster:
> > I tired this on llamacpp (well technically koboldcpp) and it worked fine for an http api endpoint however an https one doesn't seem to work over LAN.
> 
> Did you see an error in the console or network logs?

No, I don't think it's even hitting the servers if there's a https at the front. Might be self-generated cert vs an actual signed cert issue?

DeclineThyself:
> No, I don't think it's even hitting the servers if there's a https at the front. Might be self-generated cert vs an actual signed cert issue?

I have little experience with self-signed certs, I don't think I can help you here.
If I was debugging this, I'd try routing the request through mitmproxy and add a `console.log` in`generateResponse`.

Good luck!

PR #85, opened 2026-02-08
Feat/separate image profiles

munimunigamer:


src/lib/services/ai/image/InlineImageService.ts
```diff
@@ -111,7 +107,7 @@ export class InlineImageGenerationService {
 
     // Determine which profile and model to use
     let profileId = imageSettings.profileId
-    let modelToUse = imageSettings.model
+    let modelToUse = settings.getImageProfile(profileId ?? '')?.model ?? ''
```
a-frazier, line 113: This pattern is repeated quite a bit.  Maybe make a helper function for this?

.github/workflows/lint-and-typecheck.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - develop
```
a-frazier: Given `develop` doesn't exist, probably dump this line altogether.

PR #92, opened 2026-02-09
Feat/jinja prompting

munimunigamer:


LICENSE.md
a-frazier, line 1: Why are there changes to the license? (They look minor, but the file should be untouched.)

a-frazier (changes requested):
Wrong branch?

src-tauri/migrations/025_preset_packs.sql
```diff
@@ -0,0 +1,65 @@
...
+  'Aventura',
+  1,
+  strftime('%s', 'now') * 1000,
+  strftime('%s', 'now') * 1000
+);
+
+-- Assign default pack to all existing stories
+UPDATE stories SET pack_id = 'default-pack' WHERE pack_id IS NULL;
+
+-- Performance indexes
```
a-frazier, line 62: Does sqlite not create indexes on pks by default?

a-frazier (approved):

src-tauri/migrations/025_preset_packs.sql
a-frazier: Never mind, I read those wrong.

PR #144, opened 2026-02-15
feat: Reverse-connection sync model (mobile as server, PC as client)

MrLinks75:
Refactors local network sync to invert the connection topology, solving the PC firewall problem:

- **Mobile as server** on fixed port 55555, PC initiates outbound connections — bypasses PC firewalls (Windows Defender, ESET, etc.)
- **Three discovery methods**: QR code (primary), subnet-directed UDP auto-discovery, manual IP + 6-digit code
- **Real-time mobile feedback**: Activity log shows when PC connects, pulls, or pushes
- **Robust conflict handling**: Non-fatal backups/deletes so sync works even with leftover artifacts after reinstall

Tested: QR sync, manual entry, auto-discovery, push/pull both directions, firewall enabled on Windows 11 PC — all working.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -282,10 +282,11 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aventura"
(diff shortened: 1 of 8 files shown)


a-frazier:
Closed at submitter's request.

PR #174, opened 2026-02-28
Add missing kotlin bridge file

Javedz678:


src-tauri/gen/android/app/src/main/java/com/karelian/aventura/GenerationForegroundService.kt
```diff
@@ -0,0 +1,118 @@
...
+            .setForegroundServiceBehavior(NotificationCompat.FOREGROUND_SERVICE_IMMEDIATE)
+            .build()
+    }
+
+    // -- Wake Lock -------------------------------------------------------------
+
+    private fun acquireWakeLock() {
+        if (wakeLock == null) {
+            val pm = getSystemService(Context.POWER_SERVICE) as PowerManager
+            wakeLock = pm.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, WAKE_LOCK_TAG)
```
a-frazier: I don't know Android well enough to say one way or another, but should the WakeLock object just have its lifecycle tied to the service itself?  Created and Destroyed as the service is created and destroyed?  And acquire/release be tied to service start and stop?  it kind of looks like the concepts are getting mixed here.

Javedz678: Yes you are right. I will clean up the WakeLock bit.

PR #252, opened 2026-03-11
Fix Circular Dependencies: Part 1

a-frazier:
Extract logging to standalone component
Move parseImageSize to src/lib/utils/image

### DIFF
--- a/src/lib/log.ts
+++ b/src/lib/log.ts
@@ -26,8 +26,3 @@ export function createLogger(serviceName: string) {
     }
   }
 }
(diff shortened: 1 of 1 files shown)


src/lib/log.ts
a-frazier, line 1: I deliberately picked `$lib/log` for logging because it's one of very few services that's extremely cross-cutting in a way that few others are.  (Service Location / DI is one of the others.)  Pretty much everything else can be sanely namespaced.

PR #261, opened 2026-03-11
fix: vault visual descriptors manual entry

failerko:


src/lib/utils/visualDescriptors.ts
```diff
@@ -49,7 +49,8 @@ export function stringToDescriptors(input: string): VisualDescriptors {
   if (!input.trim()) return {}
 
   const result: VisualDescriptors = {}
-  const categoryPattern = /\b(Face|Hair|Eyes|Build|Clothing|Accessories|Distinguishing):\s*/gi
+  const categoryPattern =
+    /\b(Face|Hair|Eyes|Build|Clothing|Accessories|Distinguishing features):\s*/gi
```
a-frazier: Expanding `Distinguishing` to `Distinguishing features`? I don't mind, it just makes me twitch a bit since it's the only category with two words. No synchronization needed in the services or DB?

a-frazier (approved):

PR #282, opened 2026-03-22
Feat/vault assistant improvements

munimunigamer:
Just general improvements to the vault assistant from bugs/stuff I noticed using it

## Summary by CodeRabbit

- **New Features**
  - Streaming assistant replies now update live with text, reasoning, tool activity, and a dedicated preview of pending changes.
  - Added a stop-generation flow with consistent Escape/stop behavior, plus improved generation UI (including a blinking cursor).
  - Tools can load on demand by category for more focused interactions.
  - Conversation titles can be renamed directly from the selector.

- **Bug Fixes**
  - Improved persistence so concurrent/rapid generation actions don’t overwrite each other.
  - Preserved in-progress edits when new pending changes arrive.
  - Prevented duplicate vault values when composing multiple changes.
  - More reliable aborted vs error handling during streaming.


a-frazier (approved):

Pento95:
Rebased onto current `master` (this branch was 4 commits behind, including the #310 mobile tab-system restructure that touched `InteractiveVaultAssistant.svelte` — merged, then re-ported the UI-dependent pieces by hand rather than a mechanical conflict resolve, since a naive merge would have silently dropped one side or the other in a few spots).

Changes added on top of the original PR:

- **Save race fix**: `handleApprove`/`handleReject`/`handleApproveAll`/`handleNewConversation`/`handleSwitchConversation` were calling `saveConversation()` directly, bypassing the `pendingSave` queue used by `handleSend`. A save in flight from one of these could race a queued save from another and, worse, run against a `service` instance that had just been replaced (new conversation/switch), risking duplicate conversation rows. All save call sites now go through the same queue.
- **Clean abort**: stopping generation (Stop button or Escape) surfaced as a generic "Sorry, I encountered an error: ..." bubble instead of stopping silently. The service's catch block didn't special-case `AbortError`, and in Tauri's WebKit-based webview the thrown value on abort doesn't reliably satisfy `instanceof Error` either, so the check missed it either way. Added a dedicated `aborted` stream event, keyed off `signal.aborted` rather than the shape of the caught error.
- **Stop now clears the in-progress message** instead of leaving a frozen, cursor-less partial response in the chat.
- **Tool-only step feedback**: a step where the model calls a tool with no preceding text/reasoning (e.g. `list_characters`) showed no activity indicator at all until the step finished. `tool_start` now also creates the streaming placeholder so the tool chip has somewhere to render immediately.
- **Fixed a pre-existing, unrelated bug found during manual testing on this branch**: proposing more than one `update_character`/`update_scenario` edit for the same entity before either was approved caused tags/traits to duplicate on every additional proposal (visibly, e.g. 4 tags → 8 → 12), and made the tag editor effectively uneditable (local edits got silently discarded whenever the next AI response arrived). Root cause was in the pending-change composition logic (`vaultMerge.ts`'s array-diff step didn't dedupe against the already-composed base) plus a missing dirty-state guard in `VaultEntityEditPanel`'s edit-mode sync effect. Not part of this PR's original scope, but shipping it alongside since it was found while testing this branch.
- Added test coverage for the service layer — dynamic tool loading, save/load round-trip, and the abort-vs-error distinction — none of which had tests before. 87/87 passing, `svelte-check` clean.
- Minor: `stopWhenDone` now imports directly from `stopConditions` instead of the `agents` barrel, which was pulling in the whole factory/settings dependency chain for one pure function.

Pento95:
Addressed the CodeRabbit findings:

- `queueSave()` now reports success/failure so a failed save blocks the destructive transition (new conversation / switch) instead of being silently discarded.
- Rename/delete buttons in the conversation list are no longer hover-only (matched the existing `sm:opacity-0`/`focus-visible` pattern from `BranchPanel.svelte`).
- Generation ownership is now identity-guarded, so a stale `handleSend` resolving after abort/a new generation can't clobber newer state.
- Added the missing `abort` case for the SDK's native fullStream part (previously fell through as a `done`).
- Fixed the appended-item dedup in `vaultMerge.ts` to not eat a genuinely-intended duplicate.
- Added test coverage for all of the above.

Skipped two suggestions with reasons in the commit message: the Enter-during-generation one (already blocked by the disabled textarea) and the field-level merge refactor (bigger change than the bug needs; the existing whole-form guard already fixes it).

Also found and fixed one more: `handleDeleteConversation` wasn't aborting in-flight generation before resetting the service when deleting the active conversation.

PR #291, opened 2026-03-28
Feature/#258/story context unification

failerko:


src/lib/components/settings/tabs/story-settings.svelte
```diff
@@ -14,22 +13,22 @@
   </div>
 
   <WritingStyleFields
-    selectedPOV={storySettings.pov ?? 'second'}
-    selectedTense={storySettings.tense ?? 'present'}
-    tone={storySettings.tone ?? ''}
-    visualProseMode={storySettings.visualProseMode ?? false}
+    selectedPOV={story.settings.pov}
+    selectedTense={story.settings.tense}
```
a-frazier, line 17: Are these no longer nullable?  (Yet the other items are?)

src/lib/services/ai/generation/ClassifierService.ts
```diff
@@ -96,50 +77,16 @@ export class ClassifierService extends BaseAIService {
...
-      characters,
-      locations,
-      items,
-      storyBeats,
-      storyBeatTypes: 'milestone, quest, revelation, event, plot_point',
-      itemLocationOptions: 'inventory, worn, ground, or specific location name',
-      defaultItemLocation: 'inventory',
-      sceneLocationDesc: 'Name of current location if identifiable, null otherwise',
+    log('classify', {
+      narrativeLength: (story.generationContext.narrativeResult?.content ?? '').length,
```
a-frazier, line 81: This isn't an expensive fetch, is it?  (Guessing it's just a prop read, just making sure.)

src/lib/services/ai/retrieval/EntryRetrievalService.ts
```diff
@@ -166,26 +155,33 @@ export class EntryRetrievalService extends BaseAIService {
...
       totalEntries: entries.length,
       userInputLength: userInput.length,
       recentCount: recentStoryEntries.length,
       currentPosition,
-      liveCharacters: liveState?.characters.length ?? 0,
-      liveLocations: liveState?.locations.length ?? 0,
-      liveItems: liveState?.items.length ?? 0,
+      liveCharacters: characters.length ?? 0,
+      liveLocations: locations.length ?? 0,
+      liveItems: items.length ?? 0,
```
a-frazier, line 184: Do these need to be filtered from the story at all?

Edit: The prop name is "live", but the counts aren't filtered at all, is why I'm asking.

src/lib/services/generation/phases/PostGenerationPhase.ts
```diff
@@ -103,93 +36,68 @@ export interface PostGenerationResult {
...
         }
       } else {
         try {
-          result.actionChoices = await this.generateActionChoices(input)
+          result.actionChoices = await this.generateActionChoices(settings.translationSettings)
         } catch (error) {
           yield this.errorEvent(error)
+          return
         }
       }
```
a-frazier: Dumb little nitpick: Can we push the if inside the `try` so we're not duplicating the exception handling logic?

src/lib/services/generation/ActionInputController.ts
```diff
@@ -0,0 +1,640 @@
...
+  } catch (error) {
+    log('Input translation failed (non-fatal), using original', error)
+    return { promptContent: content, originalInput: undefined }
+  }
+}
+
+const sendGenerationNotification = async (responseText: string, success: boolean) => {
+  try {
+    const { sendNotification, isPermissionGranted } =
+      await import('@tauri-apps/plugin-notification')
```
a-frazier: Should this be down here or at the top of the file?

src/lib/services/generation/ActionInputController.ts
```diff
@@ -0,0 +1,640 @@
...
+  // --------------------------------------------------------------------------
+
+  async generateResponse(options?: { countStyleReview?: boolean; styleReviewSource?: string }) {
+    const userActionContent = story.generationContext.userAction?.content ?? ''
+    const userActionEntryId = story.generationContext.userAction?.entryId ?? ''
+    const countStyleReview = options?.countStyleReview ?? true
+    const styleReviewSource =
+      options?.styleReviewSource ?? (countStyleReview ? 'new' : 'regenerate')
+
+    if (!story.isLoaded) return
```
a-frazier: Probably could make this the first line of the function.  😂

src/lib/services/generation/ActionInputController.ts
```diff
@@ -0,0 +1,640 @@
...
+    const inlineImageMode = story.settings.imageGenerationMode === 'inline'
+    const streamingEntryId = crypto.randomUUID()
+    const narrationEntryId = crypto.randomUUID()
+
+    ui.setGenerating(true)
+    ui.clearGenerationError()
+    ui.clearActionChoices(story.id!)
+    ui.startStreaming(visualProseMode, streamingEntryId)
+
+    const storyId = story.id!
```
a-frazier: Could move this up so it could be used on line 154

src/lib/services/generation/ActionInputController.ts
```diff
@@ -0,0 +1,640 @@
...
+          storyId,
+        }
+
+        handleEvent(event, eventState)
+
+        if (event.type === 'phase_complete' && event.phase === 'retrieval') {
+          ui.setLastLorebookRetrieval(
+            story.generationContext.retrievalResult?.lorebookRetrievalResult ?? null,
+          )
+        }
```
a-frazier: Could this go into `handleEvent()`, or does it lack the reference to `story`?

src/lib/services/generation/ActionInputController.ts
```diff
@@ -0,0 +1,640 @@
...
+          const translationSettings = settings.translationSettings
+          if (TranslationService.shouldTranslateWorldState(translationSettings)) {
+            const translationService = new WorldStateTranslationService()
+            translationService
+              .translateEntities()
+              .catch((err) => log('World state translation failed (non-fatal)', err))
+          }
+        }
+
+        if (event.type === 'phase_complete' && event.phase === 'translation' && narrationEntry) {
```
a-frazier, line 252: There are three blocks for `phase_complete`...  Might just be GitHub's view making it unwieldly.  It is fairly readable though.  Wondering if there's a better way to organize this.  `switch`es and `function`s?  🤷‍♂️

src/lib/services/generation/ActionInputController.ts
```diff
@@ -0,0 +1,640 @@
...
+            result: event.result,
+          })
+          await story.classification.applyClassificationResult(event.result, narrationEntry.id)
+          await story.entry.updateEntryTimeEnd(narrationEntry.id)
+
+          const translationSettings = settings.translationSettings
+          if (TranslationService.shouldTranslateWorldState(translationSettings)) {
+            const translationService = new WorldStateTranslationService()
+            translationService
+              .translateEntities()
```
a-frazier, line 247: Is this an LLM call that's being waited on?

This is in the `classification_complete` event to translate the results of the world state updates, right?

src/lib/services/generation/BackgroundTaskCoordinator.ts
```diff
@@ -57,14 +30,35 @@ export interface BackgroundTaskResult {
...
+  private chapterService = new ChapterService()
+  private loreCoordinator = new LoreManagementCoordinator()
+  private styleScheduler = new StyleReviewScheduler()
 
-  constructor(deps: BackgroundTaskDependencies) {
+  /* constructor(deps: BackgroundTaskDependencies) {
     this.chapterService = new ChapterService(deps.chapterService)
     this.loreCoordinator = new LoreManagementCoordinator(deps.loreManagement)
     this.styleScheduler = new StyleReviewScheduler(deps.styleReview)
+  } */
```
a-frazier: Delete the dead code

src/lib/services/prompts/templates/loader.ts
```diff
@@ -0,0 +1,85 @@
...
+    id: meta.id,
+    name: meta.name,
+    category: meta.category as PromptCategory,
+    description: meta.description,
+    content,
+    userContent,
+  }
+}
+
+const liquidFiles: Record<string, string> = import.meta.glob('./**/*.liquid', {
```
a-frazier, line 58: Perhaps this is going to be a dumb question: How is this going to work for end users?  How will they get the default templates?

Edit: Asked Claude about this. Now I understand. Neat!

src/lib/stores/story/branch.svelte.ts
```diff
@@ -0,0 +1,724 @@
...
+
+    const dbEntry = await database.getStoryEntry(forkEntryId)
+    if (!dbEntry) {
+      throw new Error(`Fork entry ${forkEntryId} not found in database`)
+    }
+
+    const dbStory = await database.getStory(this.story.id!)
+    if (!dbStory) {
+      throw new Error(`Story ${this.story.id!} not found in database`)
+    }
```
a-frazier: This isn't being used.  Any reason to be hitting the DB again?  Can we rely on foreign keys to expect that the story exists in the DB?

src/lib/stores/story/branch.svelte.ts
```diff
@@ -0,0 +1,724 @@
...
+      const entries = await database.getStoryEntriesForBranch(
+        this.story.id!,
+        branch.id,
+        childForkPosition ?? undefined,
+      )
+      count += entries.length
+    }
+
+    return count
+  }
```
a-frazier, line 686: The SQL abuser in me thinks this can be done in pure SQL, but we don't need to worry about that right now.  Should probably put a pin in that for later though.

src/lib/stores/story/chapter.svelte.ts
```diff
@@ -0,0 +1,305 @@
...
+  private _cachedLastChapterEndIndex: number = 0
+  private _lastChapterEndIndexDirty: boolean = true
+  private _lastChaptersLength: number = 0
+  private _lastEntriesLength: number = 0
+
+  // Add a chapter
+  async addChapter(chapter: Chapter): Promise<void> {
+    if (!this.story.id) throw new Error('No story loaded')
+
+    await database.addChapter(chapter)
```
a-frazier, line 27: ... no validation of `chapter` at all?

src/lib/stores/story/chapter.svelte.ts
```diff
@@ -0,0 +1,305 @@
...
+      return 1
+    }
+
+    // Find the maximum chapter number in the current branch view
+    const maxNumber = Math.max(...this.chapters.map((ch) => ch.number))
+    return maxNumber + 1
+  }
+
+  // Update a chapter's summary
+  async updateChapterSummary(chapterId: string, summary: string): Promise<void> {
```
a-frazier: This seems like it could either forward to `updateChapter` or just not exist.

src/lib/stores/story/chapter.svelte.ts
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    // Get the entries for this chapter
+    const chapterEntries = this.story.entry.rawEntries.slice(startIndex, endEntryIndex)
+    if (chapterEntries.length === 0) {
+      throw new Error('No entries to create chapter from')
+    }
+
+    // Import aiService dynamically to avoid circular dependency
+    const { aiService } = await import('$lib/services/ai')
```
a-frazier, line 104: 🤯 

This feels like cheating.

src/lib/stores/story/chapter.svelte.ts
```diff
@@ -0,0 +1,305 @@
...
+        (ch) => !chaptersToDelete.includes(ch.id),
+      )
+      // Invalidate chapter cache after deletions
+      this.invalidateChapterCache()
+    }
+
+    // Ensure chapters are sorted by number
+    this.story.chapter.chapters = [...this.story.chapter.chapters].sort(
+      (a, b) => a.number - b.number,
+    )
```
a-frazier: Why not just sort `chapters` in place (as `sort` does) on line 165 and leave it?  `filter` won't reorder the array on line 196 so the resulting array will still be sorted.

src/lib/stores/story/chapter.svelte.ts
```diff
@@ -0,0 +1,305 @@
...
+    if (branchChapters.length === 0) return 0
+
+    // Rebuild index map if needed
+    if (this.story.entry._entryIdToIndex.size !== this.story.entry.rawEntries.length) {
+      this.story.entry.rebuildEntryIdIndex()
+    }
+
+    // Sort chapters by number to ensure we get the actual last chapter for this branch
+    const sortedChapters = [...branchChapters].sort((a, b) => a.number - b.number)
+    const lastChapter = sortedChapters[sortedChapters.length - 1]
```
a-frazier:
```suggestion
    const lastChapter = branchChapters.reduce((max, ch) => (ch.number > max.number ? ch : max))
```

src/lib/stores/story/chapter.svelte.ts
```diff
@@ -0,0 +1,305 @@
...
+
+    // Fallback: if endEntryId references a deleted entry, estimate based on entry counts
+    console.warn('Warning: Chapter endEntryId not found, using fallback calculation', {
+      chapterId: lastChapter.id,
+      chapterNumber: lastChapter.number,
+      endEntryId: lastChapter.endEntryId,
+    })
+
+    // Sum up all branch chapter entry counts as a fallback estimate
+    const totalChapterEntries = sortedChapters.reduce((sum, ch) => sum + ch.entryCount, 0)
```
a-frazier:
```suggestion
    const totalChapterEntries = branchChapters.reduce((sum, ch) => sum + ch.entryCount, 0)
```
(If you take the earlier suggestion)

src/lib/stores/story/checkpoint.svelte.ts
```diff
@@ -0,0 +1,87 @@
...
+   * @deprecated Checkpoint restoration is no longer supported.
+   * Use createBranchFromCheckpoint() instead to explore alternate timelines.
+   * This prevents data loss issues when restoring across branches.
+   */
+  async restoreCheckpoint(_checkpointId: string): Promise<void> {
+    throw new Error(
+      'Checkpoint restoration is no longer supported. ' +
+        'To explore alternate paths, create a new branch from a checkpoint instead.',
+    )
+  }
```
a-frazier: Unused.  Delete.

src/lib/stores/story/lorebook.svelte.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * Delete multiple lorebook entries (bulk operation).
+   */
+  async deleteLorebookEntries(ids: string[]): Promise<void> {
+    if (!this.story.id) throw new Error('No story loaded')
+
+    if (settings.experimentalFeatures.lightweightBranches) {
+      // COD: process each entry individually for correct tombstone handling
+      for (const id of ids) {
```
a-frazier, line 102: If this isn't going to go about it right, it might as well just call `deleteLorebookEntry()`.  :D

a-frazier (approved):
Most of these comments are left with the understanding that everything in `src/lib/stores/story/*` was really just moved with minimal changes.

src/lib/components/settings/tabs/story-settings.svelte
failerko: These have a fallback in the getter.

src/lib/services/ai/generation/ClassifierService.ts
failerko: Stuff in the `generationContext` generally contain results from individual agents, such as narrative in this case. It's only a storage class, LLM requests happen as usual within the GenerationPipeline. In other words, not expensive at all, just reading already generated data.

src/lib/services/ai/retrieval/EntryRetrievalService.ts
failerko: Uhh.. the naming's a bit scuffed i guess, "live" indicates there should be some "dead" or untracked entities, tho we don't have any like that i don't think. It's just a console log anyway :D

src/lib/services/generation/phases/PostGenerationPhase.ts
failerko: Valid

src/lib/services/generation/ActionInputController.ts
failerko: It is a little strange indeed, however not nonsensical perhaps 🤔  only importing on use... meh gonna move it out

src/lib/services/generation/ActionInputController.ts
failerko: yep, moved

src/lib/services/generation/ActionInputController.ts
failerko: No, it's async. And yea it's translating the classification results and updating the DB with the translated texts.

src/lib/services/generation/ActionInputController.ts
failerko: The issue with these is the fact they use some state that's local to `generateResponse()`

src/lib/stores/story/chapter.svelte.ts
failerko: The type is guarding it well enough i think

src/lib/stores/story/chapter.svelte.ts
failerko: Valid point.

src/lib/stores/story/chapter.svelte.ts
failerko: 🤷 😆

src/lib/stores/story/chapter.svelte.ts
failerko: Makes sense, this is a little paranoid 😆

src/lib/stores/story/lorebook.svelte.ts
failerko: I don't understand the entire lightweight branch concept enough to tell

src/lib/stores/story/branch.svelte.ts
failerko: That's a very good question :thinking:  probably fine to remove? Might be a leftover. It's already checking if a story is loaded, i think it's safe to assume the story didn't just disappear from the DB.

Pento95 (changes requested):
The "Generate Images" manual button serves no purpose: the automatic pipeline already handles image generation, and in master it was practically invisible anyway (gated behind `!!lastImageGenContext` which was only set during an active pipeline run).

Please remove it entirely from `ActionInput.svelte` — suggestions for the changed lines are attached as inline comments. Additionally, the button markup block (`{#if canShowManualImageGen}` wrapping the `Generate Images` button, unchanged from the base so GitHub does not allow a suggestion on it) also needs to be deleted.

src/lib/components/story/ActionInput.svelte
```diff
@@ -875,12 +175,11 @@
...
   async function handleManualImageGeneration() {
-    if (!lastImageGenContext || manualImageGenDisabled) return
-    const storySettings = story.currentStory?.settings
-    if (!storySettings || storySettings.imageGenerationMode === 'none') return
+    if (manualImageGenDisabled) return
+    if (story.settings.imageGenerationMode === 'none') return
     isManualImageGenRunning = true
     try {
-      await aiService.generateImagesForNarrative(lastImageGenContext)
+      await aiService.generateImagesForNarrative()
```
Pento95: In master this was also gated on `!!lastImageGenContext`, which was only set after a successful generation pipeline run — so the button was never visible on a loaded story. Now it shows immediately on any story with image mode set to "text only".

Since the automatic pipeline already handles image generation, consider removing this button entirely instead of restoring the old guard.

src/lib/components/story/ActionInput.svelte
```diff
@@ -26,93 +20,40 @@
...
   // ============================================================================
 
   let inputValue = $state('')
-  let actionType = $state<'do' | 'say' | 'think' | 'story' | 'free'>('do')
+  let actionType = $state<ActionType>('do')
   let isRawActionChoice = $state(false)
-  let stopRequested = false
-  let activeAbortController: AbortController | null = null
-  let lastImageGenContext = $state<ImageGenerationContext | null>(null)
   let isManualImageGenRunning = $state(false)
```
Pento95:
```suggestion
```

src/lib/components/story/ActionInput.svelte
```diff
@@ -26,93 +20,40 @@
...
 
-  const canShowManualImageGen = $derived(
-    story.currentStory?.settings?.imageGenerationMode === 'none' && !!lastImageGenContext,
-  )
+  const canShowManualImageGen = $derived(story.settings.imageGenerationMode === 'none')
 
   const manualImageGenDisabled = $derived.by(() => {
     if (ui.isGenerating || isManualImageGenRunning) return true
     return !hasRequiredCredentials()
   })
```
Pento95:
```suggestion
```

src/lib/components/story/ActionInput.svelte
```diff
@@ -889,202 +188,42 @@
   }
```
Pento95:
```suggestion
```

src/lib/stores/story/branch.svelte.ts
a-frazier: Yeah, we can probably dump that code.

failerko: Did that.

Pento95 (approved):

PR #313, opened 2026-04-20
model availability pings for OpenRouter (:free models) and NVIDIA NIM profiles, plus a 429 retry middleware for generation calls

Pento95:
Make room for a usable nvidia nim provider!

Model health pings
- New pingEnabled flag per profile (opt-in, only shown for OpenRouter / NIM).
- Periodic low-cost chat/completions probe (max_tokens: 1) classifies each model as ok / slow / down / auth / rate_limited, with latency and quota headers captured when available.
- Results persisted in new model_health_cache table (migration 034), keyed by (provider_id, model_id, api_key_hash). API key hashed to 8-byte SHA-256 prefix.
- Reactive SvelteMap-based store with DB hydration, concurrency-limited batch pings (5 parallel), provider-specific TTLs (auto vs manual trigger).
- Warmup on app start: pings only models actually referenced by main narrative + generation presets, deduped by provider|baseUrl|apiKey.
- UI: health indicator (signal icon + latency) inline in ModelSelector; ProfileWarningBanner blocks sending when the main narrative model is down or auth-failed.

429 retry middleware
- Wraps both doGenerate and doStream with up to 3 retries (10s / 20s / 30s + jitter).
- Respects Retry-After (delta-seconds or HTTP-date), caps at 60s to fail fast on long cooldowns.
- Abort-aware sleep: user cancellation interrupts the backoff immediately.
- Note: only catches pre-stream 429s (rejected doStream() promise), not mid-stream errors.

Other
- NVIDIA NIM profiles now ship with a default hiddenModels list to hide embedding / guard / reward / vision-only / legacy models from the selector.
- DatabaseService.init guarded against concurrent double-open via a shared pending promise.
- updateProfile purges the health cache when apiKey or providerType changes (prevents orphan rows under stale hashes).

Files
- Migration: src-tauri/migrations/034_model_health_cache.sql, registered in src-tauri/src/lib.rs
- New: src/lib/services/modelHealth{Cache,Orchestrator}.ts, src/lib/stores/modelHealth.svelte.ts, src/lib/services/ai/sdk/{middleware/retryMiddleware,providers/modelPing}.ts, src/lib/constants/modelHealth.ts,
src/lib/utils/hashApiKey.ts, src/lib/components/settings/HealthIndicator.svelte
- Touched: generate.ts, providers/config.ts, database.ts, settings.svelte.ts, ModelSelector.svelte, ProfileForm.svelte, AgentProfiles.svelte, tabs/api-connection.svelte, ProfileWarningBanner.svelte, +page.svelte


Pento95:
still testing 429 with a few providers. help with tests is welcome

src-tauri/migrations/034_model_health_cache.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS model_health_cache (
+  provider_id   TEXT NOT NULL,
+  model_id      TEXT NOT NULL,
+  api_key_hash  TEXT NOT NULL,
```
a-frazier: Is `api_key_hash` the right thing to PK on here?  I'd pick `endpoint` over that.  I don't think rotating my OpenRouter key should invalidate the model health cache, but different "OpenAI Compatible" providers definitely should have different caches based on endpoint.

src/lib/services/ai/sdk/middleware/retryMiddleware.ts
```diff
@@ -0,0 +1,111 @@
+import type { LanguageModelV3Middleware } from '@ai-sdk/provider'
+
+const RETRY_DELAYS_MS = [10_000, 20_000, 30_000]
+const JITTER_FACTOR = 0.1
+// If the provider's Retry-After tells us to wait longer than this, fail fast
+// instead of blocking the user for minutes/hours of useless retries.
+const MAX_RETRY_AFTER_MS = 60_000
```
a-frazier: Should this be linked to the user's configured timeout instead of a hardcoded value?  (Again, if it works? 😂)

Can be punted until later.  This is a good first implementation.

src/lib/services/ai/sdk/middleware/retryMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+    'statusCode' in error &&
+    (error as { statusCode: number }).statusCode === 429
+  )
+}
+
+/**
+ * Parse `Retry-After` (delta-seconds or HTTP-date) from an AI SDK error's
+ * responseHeaders. Returns null if missing or unparseable.
+ */
+function parseRetryAfterMs(error: unknown): number | null {
```
a-frazier, line 47: 💜

src/lib/services/ai/sdk/middleware/retryMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+          `[retryMiddleware] 429 with Retry-After=${Math.round(retryAfterMs / 1000)}s exceeds cap (${MAX_RETRY_AFTER_MS / 1000}s), failing fast`,
+        )
+        throw error
+      }
+
+      lastError = error
+      // Jitter only the default backoff to spread reconnections (anti-thundering-
+      // herd). Server-provided Retry-After is a floor — subtracting jitter would
+      // retry before the server is ready and earn another 429.
+      const base = retryAfterMs ?? RETRY_DELAYS_MS[attempt]
```
a-frazier, line 96: There's nothing that handles the case where retryAfterMs returns 0 due to Math.max().  I don't know if it's something that occurs in practice, but probably should be.

(It should probably be normalized to `null` so we get the backoff and jitter case.)

src/lib/services/database.ts
```diff
@@ -202,6 +218,88 @@ class DatabaseService {
...
+      latencyMs: number | null
+      quotaPercent: number | null
+      checkedAt: number
+    }>,
+  ): Promise<void> {
+    if (rows.length === 0) return
+    const db = await this.getDb()
+    // Chunked to stay below SQLite's SQLITE_LIMIT_VARIABLE_NUMBER (default 999):
+    // 8 params/row × 50 rows = 400 params per statement.
+    const BATCH_SIZE = 50
```
a-frazier: Modern versions of SQLite support 32766 variables per statement, so you can increase `BATCH_SIZE` by quite a bit... https://sqlite.org/limits.html#max_variable_number


```suggestion
    const BATCH_SIZE = 1000
```

src/lib/services/ai/sdk/middleware/retryMiddleware.ts
Pento95: i'll use the value from the slider

PR #349, opened 2026-06-13
M2.2 story entries store, actions

### LINKED ISSUE #336: Slice 2.2 — Entry-arm completion: delete, content-update, opening invariants, rollback action
**Slice doc (canonical):** [02-entry-arms.md](https://github.com/AventurasTeam/Aventuras/blob/main/docs/implementation/milestones/02-first-user-loop/slices/02-entry-arms.md) · [milestone.md](https://github.com/AventurasTeam/Aventuras/blob/main/docs/implementation/milestones/02-first-user-loop/milestone.md)

## Goal

Complete the `story_entries` Tier-1 surface: the **delete arm**
(full-row `undo_payload` capture), the **content-update
side-channel** (deliberately delta-exempt), the **opening-entry
position invariants** the generic CRUD doesn't encode, and the
**rollback-to-entry action** (suffix selection, bucket counts for
the confirm modal, execution via reverse-replay).

## Dependencies

**Depends on:** none
**Blocks:** #339 (Slice 2.5)

---

failerko:
## Summary by CodeRabbit

* **New Features**
  * Added story-entry content editing, deletion, and “rollback to a prior entry,” returning counts of affected items.
* **Enhancements**
  * More structured error responses via rejection codes.
  * Edits are blocked during in-flight generation and while rollback/reversal is in progress.
  * Improved delta reverse-and-prune behavior and working-set synchronization for entry updates.
* **Tests**
  * Expanded coverage for entry updates, deletions, rollback counts/semantics, and store hydration/patching.
* **Documentation**
  * Updated milestone/roadmap guidance for working-set store ownership and rollback semantics.


lib/actions/story-entries/operational.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+function countBuckets(rows: Pick<Delta, 'op' | 'targetTable'>[]): RollbackCounts {
+  let entries = 0
+  let chapters = 0
+  for (const r of rows) {
+    if (r.op === 'create' && r.targetTable === 'story_entries') entries++
+    else if (r.op === 'create' && r.targetTable === 'chapters') chapters++
+  }
+  return { entries, chapters, worldStateChanges: rows.length - entries - chapters }
```
Pento95, line 118: Minor, and possibly intended — worth a quick sanity check. `worldStateChanges = rows.length - entries - chapters` folds in any non-create `story_entries` delta (a metadata-update, or a delete) that lands in the window, counting it as a world-state change. The canonical bucket ([rollback-confirm.md → Counts](../../docs/ui/screens/reader-composer/rollback-confirm/rollback-confirm.md)) lists world-state as "entities, lore, threads, happenings, entry_assets" — `story_entries` isn't in it, so the consent-modal count could read slightly high. Latent in M2 (entries are created with metadata inline, so such deltas rarely fall in range). The spec is a bit ambiguous between "All other" and the enumerated list, so this may be a deliberate reading — if so, ignore me; otherwise narrowing to the named tables would tighten it.

Orthogonal to the bot threads. FWIW I think CodeRabbit's note on `reverse-replay.ts:114` (transaction commits before the patch loop, so a post-commit patcher throw gets misreported as a rollback failure) is the more important one to address — it's a real error-attribution gap, not just a count.

failerko: The naming of the field as `worldStateChanges` might be a little misleading maybe, but the "All other" does include any delta to be reversed that doesn't fall into the two higher order buckets. Technically speaking those ALSO are world state even if the data is attached to story entries.

For the rabbit finding, I was checking them at the time and since it wasn't fixed i assume it was deemed a non-issue, will re-check

PR #353, opened 2026-06-15
Slice/m2.6

### LINKED ISSUE #340: Slice 2.6 — Pack-template / Liquid engine + bundled pack
**Slice doc (canonical):** [06-pack-engine.md](https://github.com/AventurasTeam/Aventuras/blob/main/docs/implementation/milestones/02-first-user-loop/slices/06-pack-engine.md) · [milestone.md](https://github.com/AventurasTeam/Aventuras/blob/main/docs/implementation/milestones/02-first-user-loop/milestone.md)

## Goal

The prompt-template runtime: a minimal LiquidJS engine with macro
resolution, variable binding, and the include-compatibility
validator at pack load; the **bundled pack** covering every
template the M2 loop invokes; and the structural-floor
`active+in-scene` injection invariant test. This is the engine's
first landing — memory templates (M3.4), chapter-close templates
(M5.2), and the pack editor surface (M7.2) extend it.

## Dependencies

**Depends on:** none
**Blocks:** #337 (Slice 2.3), #339 (Slice 2.5), #341 (Slice 2.7)

---

failerko:
## Release Notes

* **New Features**
  * Added bundled story prompt templates (including per-turn narrative and wizard flows) with conditional sections and support for structured (JSON) outputs.
  * Added a synchronous prompt rendering engine with ID validation and custom template filters.
  * New/seeded stories now default to the bundled pack automatically.

* **Tests**
  * Expanded coverage for template rendering, filters, include validation, and bundled-template structural invariants.

* **Documentation**
  * Updated milestone and implementation notes for the prompt-template contract and output formatting.

* **Chores**
  * Added the LiquidJS dependency.


lib/prompts/engine.ts
```diff
@@ -0,0 +1,26 @@
...
+// and macro sources keyed by id, so `{% include 'id' %}` resolves from the pack.
+export function createEngine(pack: Pack): Liquid {
+  const templates: Record<string, string> = {}
+  for (const [id, entry] of Object.entries(pack.templates)) templates[id] = entry.source
+  for (const [id, entry] of Object.entries(pack.macros)) templates[id] = entry.source
+
+  const engine = new Liquid({ templates })
+  registerFilters(engine)
+  return engine
+}
```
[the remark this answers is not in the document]
failerko: templates and macros can not be undefined or null, it's typed as record always.

Pento95 (commented):
Two findings not already raised by the bots (the collision throw, jsonFilter try/catch, and filter null-guards are already addressed in the current code). Both are minor and not triggered by the M2 fixtures, but the templates/validator here are the durable artifacts 2.3/2.5/2.7 build on.

lib/prompts/bundled/per-turn.ts
```diff
@@ -0,0 +1,38 @@
...
+## {{ e.name }}
+{{ e.description }}
+{% endif -%}
+{%- endfor %}
+
+{% endif -%}
+# Story so far
+{% for entry in entries %}
+{{ entry.content }}
+{% endfor %}
```
Pento95, line 37: **Guard `# Story so far` against an empty entry buffer.** The genre/tone/setting sections have empty-guards, but this header always renders. With an empty/absent `entries` (a fresh branch, or a rollback to zero entries) it leaks a dangling `# Story so far` — exactly what the "no dangling section headers" acceptance criterion targets. Not hit by the M2 fixtures (entries is always non-empty there), but the template is what 2.7+ render through.

```suggestion
{% if entries.size > 0 -%}
# Story so far
{% for entry in entries %}
{{ entry.content }}
{% endfor %}
{% endif -%}
```

Note: applying this shifts whitespace slightly, so regenerate the snapshot afterwards: `pnpm test:run -u lib/prompts/bundled/per-turn.test.ts`.

lib/prompts/validate-includes.ts
```diff
@@ -0,0 +1,40 @@
...
+export function extractIncludes(source: string): string[] {
+  const ids = new Set<string>()
+  for (const match of source.matchAll(INCLUDE_RE)) ids.add(match[1])
+  return [...ids]
+}
+
+// A template in group G may include only G-tagged or `staticContent` macros.
+export function validatePackIncludes(pack: Pack): IncludeViolation[] {
+  const violations: IncludeViolation[] = []
+  for (const [templateId, template] of Object.entries(pack.templates)) {
```
Pento95, line 16: **The validator only checks `pack.templates` includes, never macro→macro.** The invariant "a member of group G may include only G-tagged or `staticContent` macros" isn't enforced when a *macro* includes another macro with a mismatched (or missing) group — that loop never visits `pack.macros` sources. Inert for M2 (the bundled `staticContent` macros include nothing), so not a blocker. Worth either iterating over macro sources here too, or leaving an explicit forward-note toward the parked M7.2 author-time validator so the gap is tracked rather than silent.

lib/prompts/bundled/per-turn.ts
failerko: The template is just a placeholder and also empty `entries` is an unreachable state under normal circumstances. As currently designed, all stories have an opening entry created in wizard which is undeletable.

lib/prompts/validate-includes.ts
failerko: Fair enough, added a note to M7.2's roadmap section so it doesn't get forgotten and is included properly in that slice.

PR #363, opened 2026-07-05
Slice 2.5 — Reader-composer minimum

### LINKED ISSUE #339: Slice 2.5 — Reader-composer minimum
**Slice doc (canonical):** [05-reader.md](https://github.com/AventurasTeam/Aventuras/blob/main/docs/implementation/milestones/02-first-user-loop/slices/05-reader.md) · [milestone.md](https://github.com/AventurasTeam/Aventuras/blob/main/docs/implementation/milestones/02-first-user-loop/milestone.md)

## Goal

The core screen, minimum-viable: virtualized entry window with
load-older pagination and scroll-anchoring on prepend, autoscroll
and jump buttons, the composer with send-time mode wrapping and
send→cancel duality, trigger generation, markdown rendering
pipeline (htmlStreaming port), Harper.js spellcheck, basic edit /
delete entry actions with the **rollback-confirm modal compound**,
and CTRL-Z basic single-action undo + redo. Calendar time chrome
(top-bar chip + per-entry footer labels) renders through
`lib/calendar`.

## Dependencies

**Depends on:** #336 (Slice 2.2), #340 (Slice 2.6), #337 (Slice 2.3) — partial gates; the bulk is day-one against the stub seam
**Blocks:** #341 (Slice 2.7); #338 (Slice 2.4)'s debug-button removal task

---

Pento95:
## Summary

Reader-composer screen minimum: virtualized entry window with load-older + web scroll-anchor compensation, mode-wrapping composer (Do/Say/Think/Free), an interim real per-turn dispatch (stub/real provider, streaming, fault injection via config pre-flight), markdown/sanitization pipeline (web + native), Harper.js spellcheck, edit/delete via the rollback-confirm modal, CTRL-Z/redo, and edit-restriction gating. Calendar time chrome is gated on Slice 2.3 and not included here.

Key architectural fix found during review: `submitTurn`'s `user_action` write now shares one `actionId` with the pipeline run (milestone C6 contract), fixing a real bug where CTRL-Z reversed only the AI reply and orphaned the user's entry. `RunCtx` gained an optional caller-supplied `actionId` to make this possible; `selectUndoTarget` was corrected to anchor a turn at its earliest `story_entries` create.

Two follow-ups logged in `docs/followups.md`: `EntryCard` doesn't yet consume the markdown pipeline built here (renders plain text); jump buttons/autoscroll aren't functionally wired pending an imperative scroll API on `EntryWindow`.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean
- [x] `pnpm vitest run --project unit` — 758/770 pass (12 pre-existing, unrelated `lib/ai` SDK-version failures, confirmed present on `main` before this branch)
- [ ] Manual cross-platform matrix (prepend anchoring, keyboard avoidance on phone, Harper toggle, autoscroll during a real stream) — tracked as open in the slice doc, needs a running app pass

https://claude.ai/code/session_0151EkuaZvXs6evwZHxbuPcR

## Summary by CodeRabbit

* **New Features**
  * Added a full reader composer with optional mode selection, spellcheck/linting, and send/cancel controls.
  * Introduced a virtualized entry window with jump-to-top/bottom and a rollback confirmation flow.
  * Enabled streaming narrative generation with safer incremental HTML rendering.
* **Bug Fixes**
  * Improved reader scroll stability during streamed updates (prepend compensation/autoscroll).
  * Hardened undo/redo and turn replay behavior to be more reliable.
* **Tests**
  * Expanded coverage for streaming HTML safety, composer behavior, turn submission, and undo/redo.
* **Documentation**
  * Updated reader UX follow-ups and implementation milestone notes.


Pento95:
Went through both automated reviews (CodeRabbit + Gemini Code Assist) and verified each finding against the code independently before acting. Summary:

**Fixed (8 commits, `d4eb3606`..`1e82a81d`):**
- Global Ctrl/Cmd+Z was hijacking native undo in text inputs — now skips app-level undo/redo when the event target is editable (`b63bdac9`)
- `submitTurn` could throw synchronously with no try/catch, leaving the UI stuck with no error surfaced (`b63bdac9`)
- `commitEdit`/`confirmRollback` discarded in-progress state (draft/modal) even on a rejected result — now preserved + toast on rejection (`b63bdac9`)
- `redoLastAction` popped the redo stack before confirming `applyRedo` succeeded, and ignored `branchId` entirely (cross-branch redo corruption risk) — most serious finding in the batch, fixed with a peek-validate-apply-pop sequence + branch check (`09762a82`)
- Undo could lose the redo snapshot if the post-commit store-sync threw after the DB reversal already landed (`09762a82`)
- `stream-buffer.ts`: bare `<` in prose was withheld forever; `>` inside a quoted attribute value fooled the tag-close check — replaced with a proper forward scan tracking tag/quote state (`d4eb3606`)
- `sanitize.ts` allowed `style` in the attribute allowlist (DOMPurify doesn't vet CSS contents, e.g. `url(...)` exfiltration) — dropped from the allowlist (`f73f6e88`)
- `redo.ts` could patch a phantom empty row into the store when nothing was actually written to the DB — patcher now skipped for null-`rowBeforeUndo` create/update (`edbf206c`)
- `autoscroll.ts`'s programmatic-scroll marker wasn't cleared after a non-matching event, so a later coincidental match could wrongly swallow a real user scroll — now a true one-shot guard (`1e82a81d`)

All fixes ship with regression tests confirmed to fail against the pre-fix code.

**Not fixed — false positives, verified and dismissed:**
- Gemini's prompt-race claim on `pipeline.ts` (`entriesStore` read racing the DB write) — the patcher call is synchronous before `runPipeline` ever starts; no race exists.
- CodeRabbit's "Open questions section violates an immutable slice-doc rule" — no such rule exists anywhere in this repo's conventions; the section is explicitly meant to stay open or get resolved during implementation.
- Gemini's position-race comment on `submit-turn.ts` — already fixed upstream of that review's SHA (`e66d1668`).

**Acknowledged, deliberately deferred:**
- `commitEdit`/`confirmRollback`'s sibling risk in the reload effect (no cancellation guard on fast branch-switch) — not addressed in this pass, worth a follow-up.
- `composer-wrap`'s third-person conjugation punctuation bug — real, but currently dead code (POV is hardcoded first-person until story-settings UI lands).
- `MODE_OPTIONS` built at module load via `t()` — latent i18n-reactivity gap, but no runtime locale switching exists yet anywhere in the app.
- Minor nitpicks (missing `.limit()` on the undo delta scan, `UndoCandidateDelta.source` type narrowing, a redundant copy string) — low value, skipped for now.

Full unit suite: 771/783 passing after all fixes; the 12 failures are pre-existing `lib/ai/*` AI-SDK-version issues unrelated to this branch (confirmed identical on `main`).

app/reader-composer/[branchId].tsx
```diff
@@ -1,33 +1,311 @@
...
+    return () => {
+      cancelled = true
+    }
+  }, [branchId])
+
+  useEffect(() => {
+    if (entriesStore.getLoadedBranch() !== branchId) void reload()
+  }, [branchId, reload])
+
+  const surfaceTurnFailure = useCallback(
```
Comment (failerko) [inline app/reader-composer/[branchId].tsx]: I'd rename this to `showTurnFailure`

components/reader/composer.tsx
```diff
@@ -0,0 +1,140 @@
...
+  onSend: (rawText: string, mode: ComposerMode) => void
+  onCancel: () => void
+}
+
+const MODE_OPTIONS: SelectOption[] = [
+  { value: 'do', label: t('reader:composerMode.do') },
+  { value: 'say', label: t('reader:composerMode.say') },
+  { value: 'think', label: t('reader:composerMode.think') },
+  { value: 'free', label: t('reader:composerMode.free') },
+]
```
failerko: Fix coderabbit's comment on the `MODE_OPTIONS` i18n issue plz. It's true weonly have a single language and the issue is latent, but no need to leave landmines for when we add other laguages.

lib/markdown/sanitize.ts
```diff
@@ -0,0 +1,24 @@
...
+  'pre',
+  'ul',
+  'ol',
+  'li',
+  'br',
+  'hr',
+  'a',
+  'span',
+]
+const ALLOWED_ATTR = ['href', 'class']
```
failerko: Add `style` and `color` as well, common to see in AI output, also `font` to `ALLOWED_TAGS`, same reason.
Technically speaking `<font color="">` is old/deprecated, but i've seen it in plenty of presets.

failerko (dismissed):

failerko:
> * `composer-wrap`'s third-person conjugation punctuation bug — real, but currently dead code (POV is hardcoded first-person until story-settings UI lands).

@Pento95 check this one too plz, your comment says deliberately deferred because currently dead code. You gotta rein in your AI a little to not be lazy.
Might be dead code, but is leaving a landmine that creates a bug later.

Pento95:
@failerko
- Renamed surfaceTurnFailure → showTurnFailure.
- composer.tsx's MODE_OPTIONS now built at render time instead of module load, so it'll react to locale changes.
- sanitize.ts now allows style/color/font for AI-output formatting, but style is scoped to a property allowlist (color, font-weight, font-style, text-decoration) instead of a free-text attribute — keeps the CSS-exfiltration fix from an earlier commit intact.

CodeRabbit
- submitTurn now reverses the persisted user_action when pipeline admission is rejected (a path abortRun's usual delta-reversal never reaches, since no run was ever reserved) and the reader screen surfaces that rejection to the user instead of silently swallowing it.

Found during self-review & testing
- undoLastAction/redoLastAction now bracket their sweep with reversalInProgress/isUserEditBlocked, matching rollbackToEntry — closes a race where Ctrl+Z mid-generation could corrupt state.
- Rollback and edit now clear the redo stack too (previously only new turns did), so a stale redo can't resurrect content a rollback just deleted.
- applyRedo preserves its snapshot when a post-commit store-sync fails, instead of leaving a redo permanently stuck on a PK conflict.
- Closed a real position-collision race: two turns submitted close together on the same branch (user_action position, and the ai_reply's position inside the pipeline phase) could land on the same slot. Fixed with an in-process per-branch queue — caught by a regression test written for the fix.
- Ctrl+Z/Shift+Z, save-bar's Cmd+S, and actions-menu's Cmd+K were three independent copies of the same keydown-shortcut scaffold, none with an event.repeat guard. Extracted a shared useGlobalHotkey hook.
- conjugateThirdPersonPresent had no consonant+y→ies rule ("carrys" instead of "carries") — latent today since POV is hardcoded to first-person, but wrong the moment that changes.
- Deduped whereForDelta, copy-pasted identically into redo.ts and reverse-replay.ts.
- Dropped a comment naming slice numbers, which rots the moment the slice ships.

lib/markdown/sanitize.ts
```diff
@@ -0,0 +1,71 @@
...
+  'a',
+  'span',
+  'font',
+]
+const ALLOWED_ATTR = ['href', 'class', 'style', 'color']
+
+// `style` values are limited to this property allowlist rather than
+// blacklisting patterns like `url(` — the allowed properties' value grammar
+// has no URL/expression form, so there's no exfiltration surface to filter.
+const ALLOWED_STYLE_PROPS = new Set(['color', 'font-weight', 'font-style', 'text-decoration'])
```
failerko: This is too limiting, it only assumes font styling and completely kills any possibility of HTML+CSS visual elements. I think this pattern should be reversed, pick out the couple props that are actually a danger and strip only those.

failerko (changes requested):

lib/markdown/sanitize.ts
Pento95: Absolutely, thanks for spotting this

failerko (approved):

failerko (approved):

PR #367, opened 2026-07-15
Native SAF export/backup/restore (Android OOM fix) + export/import fixes, #330, #116

### LINKED ISSUE #116: Adjustable chapter summary length
With longer chapters, summaries are degraded or skipping over important parts or even getting the events wrong. This should be dynamic with chapter token count or override-able in settings.

### LINKED ISSUE #330: Pollinations image list missing image models
<kbd>[adv997](https://discord.com/channels/1456625039846805686/1508073644419711037/1508073644419711037)</kbd> [adv997](https://discord.com/channels/1456625039846805686/1508073644419711037/1508073644419711037)  `BOT`

When adding a profile for Pollinations in the "Images" section of the settings, the model list has only eight models in it.
If you refer to this documentation:  https://gen.pollinations.ai/docs#tag/%EF%B8%8F-image-generation
The available models are:
```
kontext, nanobanana, nanobanana-2, nanobanana-pro, seedream5, seedream, seedream-pro, gptimage, gptimage-large, gpt-image-2, flux, zimage, wan-image, wan-image-pro, qwen-image, grok-imagine, grok-imagine-pro, klein, p-image, p-image-edit, nova-canvas
```
I was looking to use seedream, for example.

---
[View on Discord](https://discord.com/channels/1456625039846805686/1508073644419711037/1508073644419711037)

Pento95:
## Apologies up front 🙏

This is a big, multi-topic PR — a "god-PR", and I'm sorry for that. The changes grew out of one investigation (Android OOM on export/backup) and pulled in several adjacent fixes I had queued locally. I know a bundle like this is harder to review than a focused branch, so I've documented each area separately below and I'm happy to split it into smaller PRs if you'd prefer — just say the word.

---

## Overview

The core of this work fixes **Android OutOfMemory crashes** during export, backup, restore and import, and makes those flows fast. It also folds in two issue fixes and a few UX/quality improvements.

### 1. Android OOM: native, streamed export / backup / restore

**Problem.** Exports and backups routed hundreds of MB (base64 images, the full SQLite DB) through the JS/IPC bridge as strings. On Android's constrained WebView heap this reliably OOM-crashed.

**Fix.** All heavy file I/O now happens natively in Rust and streams file-to-file — nothing large ever lives in the JS heap or crosses the bridge:

- `backup_database`, `export_story_avt`, `export_single_image`, `export_images_zip` write **directly** to the chosen destination.
- On **desktop** the destination is the real path from the save dialog.
- On **Android** the destination is the SAF `content://` URI from the save dialog, opened natively via the fs plugin's content-URI file descriptor (`ContentResolver`). Writes stream in ~8 KB blocks, so peak memory is a single buffer regardless of file size.
- **Restore** mirrors this: the picked `content://` backup is streamed into a temp file in the app's internal cache dir, then restored natively.
- The `.avt` exporter fills each image's base64 from SQLite on the native side, so image bytes never touch the WebView heap.

**Perf.** Backup DB compression dropped from Deflate level 6 → **level 1** (the DB is mostly already-compressed base64 PNGs, so level 1 recovers the base64 bloat at a fraction of the CPU), and the `.avt` exporter now pulls all images in **one streamed query** instead of one query per image.

### 2. Native `.avt` import — the OOM's other half

The known limitation this PR originally shipped with ("importing an extremely large `.avt` still parses JSON in the WebView") turned out to be a hard failure, not just memory pressure, and it is now fixed.

**Problem.** The Library's Import button read the picked file with an `<input type="file">` and `file.text()`. That has to materialise the whole `.avt` as a JS string, and V8 caps a string at `2^29 - 24` bytes (~512 MiB). A real 546 MB export exceeded it by 1.8% and failed with a misleading *"not a valid JSON file"* — the file was intact; the read never delivered it. No amount of RAM helps: the ceiling is structural.

**Fix.** Import now goes through the native picker into a streaming Rust reader, mirroring the exporter's split — **JS owns the structure, Rust owns the bytes**. Id remapping, insertion order and foreign keys stay in TypeScript where they are tested; only the payloads move.

Two passes, both streaming:
1. `avt_read_light` returns the JSON with every `imageData` removed (serde's `IgnoredAny` skips each base64 without allocating it). JS parses that and runs the ordinary import.
2. `avt_import_images` re-reads the file and streams each payload straight into SQLite, using the small old-image-id → new-entry-id table JS produced.

Re-reading is deliberate: a staging table would need a migration and would leak orphan rows if an import died half way. Peak memory is one image regardless of file size — measured at **16 kB vs 108 MB** for a naive parse of a 100 MB file (`cargo test --lib avt_import -- --ignored --nocapture`).

The structure and the images cannot share a transaction (the native side writes over its own connection), so a failed image pass deletes the half-imported story rather than leaving one with its pictures silently missing.

**Verified on-device:** the 546 MB `.avt` that previously failed now imports in ~1 minute.

### 3. Export/import robustness

- New `errMessage()` helper: Tauri command rejections arrive as plain strings, so the old `e instanceof Error ? … : 'Unknown error'` discarded the real (often precise) native message. Now surfaced everywhere export/import can fail.
- Export toasts now report failures instead of failing silently (the `.avt` gather step used to run outside the try/catch).

### 4. Closes #330 — Pollinations image models missing

The provider filtered out models flagged `paid_only`, hiding valid image models from the list. Filter removed so all image-capable models appear.

The "Include paid models" toggle is gone with it: the API already filters by tier server-side from the API key, so the toggle was a client-side duplicate that was inert wherever the key was actually sent (images) and misleading where it wasn't (text). The text model fetch now sends the key too, so the list reflects what the key can really generate.

### 5. Closes #116 — Adjustable chapter summary length

Chapter summary length is now configurable rather than fixed.

### 6. Generation info popover

Narration entries now have an info popover showing the model / profile / effort / timestamp used to generate them.

---

## Review fixes

From the Gemini and CodeRabbit reviews, plus bugs found while verifying their findings. Full reasoning [in this comment](https://github.com/AventurasTeam/Aventuras/pull/367#issuecomment-4996629703).

**Restore could destroy the database.** `File::create` truncated the live `aventura.db` before extraction succeeded, so a CRC error, a truncated archive or a full disk left it in pieces — and the safety copy's failure was only `eprintln!`'d, with nothing ever restoring from it. The DB is now staged and validated first, then swapped in by **renaming** the old file aside: atomic, and free in disk space, which matters when the DB is ~270 MB on a phone. Covered by 6 Rust tests, checked to fail against the old logic.

**Backups could silently omit recent data.** When `VACUUM INTO` failed, the code fell back to archiving the live `aventura.db`. The DB runs in WAL mode (verified: header bytes `02 02`), so recent commits can live in `aventura.db-wal` — the fallback could ship a stale backup, the worst failure mode for a backup, since it only surfaces at restore time. A failed snapshot is now a hard error.

**Gallery could re-create the OOM this PR fixes.** `imageDataCache` was cleared on refresh but not on story change, and had no eviction: lazy loading bounds the peak at open, not the total. It is now an LRU with a byte budget, cleared on story change. Failed loads render a Retry instead of a permanent spinner.

**Staged SAF restore temps were never deleted** despite the doc comment claiming otherwise — full copies of the backup, hundreds of MB. They are now swept on the way *in*: a cleanup after the restore would never run, because the app calls `exit(0)` on success.

### Bugs found while reviewing the reviews

- **Checkpoint snapshots kept stale chapter ids.** Checkpoints are imported before chapters, so their snapshots' remapping found nothing in the id map and silently kept the export's original ids — a rollback on an imported story would restore chapters that do not exist. Chapters are now pre-mapped.
- **Every export was stamped v1.7.0 while the importer checked for v1.8.0**, so re-importing our own file always warned that the background image would not be restored — while it was. (This versions the `.avt` *file format*; it is unrelated to the app version, which the code never said. It does now.)
- **Android file pickers dropped their filters.** The dialog plugin maps each extension to a MIME type with `MimeTypeMap.getMimeTypeFromExtension`, which returns null for anything Android does not know — including our own `avt` and the `*` wildcard — and discards it silently. `['avt', 'json', '*']` became "JSON only", grey-ing out every `.avt`; the "All Files" escape hatch never worked. Confirmed on-device for restore too: `backup.zip (1)`, the exact duplicate-name case the fallback was written for, was unselectable. Filters are now omitted on Android, where they could only lie.
- **A dead `oldToNewId.set(image.id, …)`** wrote image ids into the same map used for entry lookups two lines below — harmless today, corrupting on a collision.

### Not applied (false positives)

- *"Potential null pointer dereference on `memoryConfig`"* — `story.memoryConfig` is a store getter that coalesces to `DEFAULT_MEMORY_CONFIG`; it never returns null.
- *"Keep one source of truth for the memory configuration types"* — `services/ai/index.ts` imports those types from `$lib/types` rather than declaring them; there is one declaration of each.

---

## Refactor

`importFromContent` was a single ~490-line function holding the most delicate logic in the project (id remapping, COW `overridesId`, topological branch insertion). It is now `services/import/`, split into `validate` / `idMaps` / `structure` / `images`, with the image payloads behind a strategy — inline for sync, native for a picked file. Both entry points share one implementation of the structure.

**Adds vitest and the first frontend tests.** 33 tests, written against the import's public entry point *before* the refactor so it could be judged by "these still pass", and checked by mutating the source to confirm they fail on a regression (one of them didn't, and was fixed). Plus 14 Rust tests, three writing to a real SQLite database.

## Testing

- `npm run check`, `npm run lint`, `cargo fmt`, `cargo clippy` clean. 33 frontend tests, 14 Rust tests.
- **On-device (Android, arm64):** verified export `.avt` (~486 MB, valid JSON confirmed byte-for-byte), backup (~363 MB, well compressed), restore (app restarts cleanly), a gallery of 800+ images, Pollinations model fetch, and the 546 MB `.avt` import that previously failed.

## Known limitations

- **Sync inherits the same ceiling.** `SyncModal` receives a story as a string over the network, so there is no file to hand to Rust and a synced story over ~512 MiB would fail the same way. That is a change to the sync transport, not the import.
- The SAF direct-write assumes the picked destination's file descriptor is seekable (true for on-device storage like Downloads); exotic document providers are untested.

## Housekeeping

`src-tauri/gen/android/app/tauri.properties` is now untracked. Tauri's own generated `.gitignore` already listed it, but a file committed before a rule exists stays tracked regardless — so it sat at `versionCode=7003` while the app shipped 0.7.6, and `versionCode` is what Android compares to allow an update. Verified that `tauri android build` recreates it from `tauri.conf.json`, which is what CI runs.

Thanks for reading through all of it — and again, sorry for the size. Happy to break it up or adjust anything.


Pento95:
Thanks both. I went through every finding and verified each one against the code rather than taking it at face value. 11 applied, 2 rejected as false positives. Details below.

## Applied

**`backup.rs` — restore truncated the live DB before extraction succeeded** (gemini high / CodeRabbit critical)

Both correct, and this was the only genuinely serious one. `File::create(&target)` truncated `aventura.db` before `io::copy` ran, so a CRC error, a truncated archive or a full disk left the live database destroyed. The safety copy was worse than useless: its failure was only `eprintln!`'d, and nothing ever restored from it.

Fixed by staging the extracted DB next to the target, validating it, and only then swapping. One deviation from both suggestions: the safety copy is now a **rename**, not a `std::fs::copy`. It's atomic, instant, and costs no extra disk — the copy needed 3x the DB size (original + copy + staged), which on a ~270 MB Android DB is not a rounding error. If the second rename fails, the first is rolled back.

Also extracted the logic into `restore_db_from_zip(zip_path, target)` (plain paths, no `AppHandle`) so it's actually testable, and added 6 tests. I ran them against the *old* logic to confirm they catch the bug — the corrupt-archive case fails there with `live DB was damaged: Invalid checksum`, with the live DB overwritten by partial bytes from the new backup.

**`backup.rs` — staged SAF temps never cleaned up** (gemini medium / CodeRabbit major)

Real, and the doc comment claiming "the temp is best-effort cleaned up by the restore" was simply false — nothing removed them. These are full copies of the user's backup (~363 MB in my testing).

I went with @gemini-code-assist's approach over CodeRabbit's here: cleanup happens on the way *in*, at the top of `import_saf_to_temp`. A `finally`-equivalent in the restore command would never run in the normal case, because the app calls `exit(0)` on success — and it would still leak if the OS killed the app mid-restore. Sweeping on entry covers both.

**`backupService.ts` — no live-DB fallback when `VACUUM INTO` fails** (CodeRabbit major)

Valid, and I confirmed the premise empirically before acting: the header bytes of my `aventura.db` are `02 02`, i.e. WAL is on. Archiving the bare `.db` really could ship a backup missing the newest commits — the worst failure mode for a backup, since it only surfaces at restore time. Fallback removed; a failed `VACUUM INTO` is now a hard error.

**`GalleryTab.svelte` — unbounded base64 cache** (CodeRabbit major)

Valid, and thematically the important one on an OOM PR. `imageDataCache` was cleared in `refreshImages()` but *not* on story change, and had no eviction — lazy loading bounds the peak at open, not the total. Now an LRU with a 96 MB budget (base64 length is a good proxy for the heap it costs), cleared on story change.

**`GalleryTab.svelte` — failed images stuck on a permanent spinner** (CodeRabbit minor)

Valid. `observer.unobserve(node)` ran synchronously right after the un-awaited `ensureImageData(id)`, so it fired regardless of outcome, and the error was swallowed. Dropped the `unobserve` entirely — it was also wrong for the new LRU, since an evicted image must reload when scrolled back — and added a `failedImageIds` set rendering a Retry button in both grid and lightbox.

**`export.ts` — duplicated comment** (CodeRabbit nitpick) — real; the first copy was also stale about Android. Removed.

**`StoryEntry.svelte` — `formatStoryTime` optional chaining** (gemini medium)

Applied, though as defensive hardening rather than a demonstrated bug: `TimeTracker`'s fields are all required, and `baseTime` falls back to a complete default. The plausible path is a partial `timeTracker` from an imported `.avt` — `??` guards null, not incomplete objects. Cheap insurance, so in it goes.

**`ExperimentalSettings.svelte` — static `open` import** (CodeRabbit nitpick) — applied, cosmetic.

## Rejected

**`MemoryView.svelte` — "potential null pointer dereference on `memoryConfig`"** (CodeRabbit major)

False positive. `story.memoryConfig` is a store **getter**, not a field:

```ts
get memoryConfig(): MemoryConfig {
  return this.currentStory?.memoryConfig || DEFAULT_MEMORY_CONFIG
}
```

It never returns null. The reasoning came from `export.ts:392` (`memoryConfig: data.story.memoryConfig || null`), but that sets the field on the *Story entity*, which the getter then coalesces. `story.memoryConfig.summaryDetail` is already safe.

**"Keep one source of truth for the memory configuration types"** (CodeRabbit nitpick)

False positive. `src/lib/services/ai/index.ts` doesn't declare those types — it *imports* them from `$lib/types` (lines 49 and 57). A repo-wide grep finds exactly one declaration of each, in `src/lib/types/index.ts`. There is no parallel contract.

## Notes

CodeRabbit's `backupService.ts:147-156` ("app left with its DB connection closed") is diagnostically wrong — `close()` nulls the handle and `getDb()` re-opens via `init()`, as documented at `database.ts:140-141`. No reopen/relaunch needed. The real risk was reopening onto a DB that the failed restore had already truncated, which the staging fix resolves at the source.

Both bots reviewed 6dba668, which predates the removal of the Pollinations "Include paid models" toggle — CodeRabbit's release notes still advertise it. The API filters by tier server-side from the key, so the toggle was a client-side duplicate that was inert wherever the key was actually sent. The list now reflects what the key can really generate.

`cargo test` 6/6, `svelte-check` clean, `cargo fmt` clean.

failerko (approved):

PR #392, opened 2026-07-22
Slice M3.2 - piggyback layer

### LINKED ISSUE #378: Slice 3.2 — Piggyback layer: trailing block, scene metadata, computed bookkeeping
**Slice doc (canonical):** [02-piggyback.md](https://github.com/AventurasTeam/Aventuras/blob/main/docs/implementation/milestones/03-memory-floor/slices/02-piggyback.md) · [milestone.md](https://github.com/AventurasTeam/Aventuras/blob/main/docs/implementation/milestones/03-memory-floor/milestone.md)

## Goal

The per-turn fast path of the memory cadence: the narrative model
emits a tagged `<state>` block alongside its prose, and this slice
parses it, applies the scene-metadata and state writes, derives the
computed bookkeeping, auto-promotes staged entities on ID emission,
and falls back to a per-turn classifier pass when the model can't
be trusted with tagged blocks. Also extracts the shared
entry-metadata placeholder helper flagged during M2 reconciliation.

## Dependencies

**Depends on:** none — day-one startable
**Blocks:** #383 (Slice 3.7)

---

---
[View on Discord](https://discord.com/channels/1456625039846805686/1529477899483218112/1529477899483218112)

Pento95:
## Summary

Implements **Slice 3.2 — Piggyback layer** ([`docs/implementation/milestones/03-memory-floor/slices/02-piggyback.md`](file:///home/pento/Aventuras/docs/implementation/milestones/03-memory-floor/slices/02-piggyback.md)), providing the per-turn fast path of the memory cadence. The narrative model emits a tagged `<state>` block alongside prose, which is parsed and applied to entry metadata, entity states, and computed bookkeeping, with a synchronous per-turn fallback classifier when tagged blocks are unreliable or fail to parse.

### Delivered Scope
- **Trailing-block parser (C2)**: Robust segment isolation for `<state>` blocks (`sceneEntities`, `currentLocation`, `worldTimeDelta`, `visualChanges`, `transfers`, `summary`) with `jsonrepair` fallback and bracketed placeholder substitution (`c1`, `l1` -> entity UUIDs).
- **Prompt fragment**: `<state>` emission instructions, staged-entity prompt framing, and calendar vocabulary (tier names, weekdays, eras) in per-turn prompt template.
- **Piggyback apply & narrow actions**: Writes `sceneEntities`, `currentLocationId`, and `worldTime` delta to entry metadata; applies visual state updates and item/stackable transfers; filters out non-existent entity IDs to enforce "piggyback creates no rows" and prevent action rejection crashes.
- **Computed bookkeeping**: Automates `current_location_id` tracking for characters in scene and `lastSeenAt` transitions upon scene exit.
- **Staged-to-active fast path**: Auto-promotes staged entities when their ID is first emitted in `<scene_entities>`.
- **Capability gate & Fallback Classifier**: Adds `piggybackFallbackClassifierPhase` to run a synchronous classifier pass when model capability is uncurated/unreliable or when trailing block parsing fails.
- **`<summary>` enrichment hand-off**: Extracts optional trailing summary for next turn's Q2 structural digest.
- **UI Presentation**: Updates `EntryCard` to strip raw `<state>` XML blocks from narrative text while providing a `Globe` header action icon to toggle an expandable preview of state metadata.

## Test Plan
- [x] `pnpm typecheck` passed with 0 TypeScript errors.
- [x] Unit test suite for `lib/piggyback/` (31/31 tests passing).
- [x] Pipeline integration tests for `per-turn-piggyback` (17/17 tests passing).
- [x] Fault scenario & error logging tests passing.
- [x] Runtime validation against SQLite DB: verified entry metadata, deltas, and entity tracking.

## Summary by CodeRabbit

* **New Features**
  * Added structured “world state” extraction with per-turn piggyback scene updates (entities, locations, world-time deltas, visual changes, item/stackable transfers), including a classifier-based fallback.
  * Updated per-turn prompts and context (staged characters, known locations, optional calendar vocabulary, piggyback firing controls).
  * Entry cards now support an optional, toggleable world-state panel.

* **Bug Fixes**
  * Delta log now renders correct labels for additional delta sources.
  * Improved entity state patching/merge behavior and staged-promotion concurrency.
  * Updated Electron security policy to allow `data:`/`blob:` connections and refined spellcheck lint suppression.


components/compounds/delta-log-row.tsx
```diff
@@ -10,6 +10,7 @@ type DeltaOp = 'create' | 'update' | 'delete'
 // is frozen, so the two won't drift.
 type DeltaSource =
   | 'ai_classifier'
+  | 'per_turn_classifier'
```
failerko, line 14: `ai_classifier` is the piggyback correct? I'd rename it to make that clear since we're tagging the dedicated per-turn verbosely as well. I kind of expected both of them to use `ai_classifier` as they do the same thing, but this is fine too, just rename to make the distinction clear.

lib/pipeline/definitions/per-turn-piggyback.ts
```diff
@@ -0,0 +1,223 @@
...
+  const entityList = referenceable
+    .map(
+      (e) =>
+        `- [${idMap.allocate(e.id)}] ${e.name} (${e.kind}${e.status === 'staged' ? ', staged' : ''})`,
+    )
+    .join('\n')
+
+  const appSettings = appSettingsStore.getAppSettings()
+  const result = await generateClassifierState(
+    `Known entities, referenced only by the bracketed ID shown below — never invent one:\n${entityList || '(none)'}\n\nExtract scene state from this reply:\n\n${tail.content}`,
```
failerko: Hardcoded prompt and all data formatting in-code?
The prompt should be external in `lib/prompts/bundled` and it should utilize `buildGenerationContext` into `renderTemplate` (same applies for any story-related prompt) and it should also handle all data formatting, in this case building entryList, in the Liquid template.
Check `lib/pipeline/definitions/per-turn.ts`, it uses this pattern.

failerko (changes requested):

lib/prompts/bundled/per-turn.ts
```diff
@@ -22,18 +22,44 @@ export const PER_TURN_NARRATIVE = `{% if definition.setting != blank -%}
...
+
 {% endif -%}
 # Story so far
 {%- assign recentEntries = entries | recent: userSettings.partialChapterBuffer %}
 {% for entry in recentEntries %}
 {{ entry.content }}
 {% endfor %}
-{% include 'macro_output_format_narrative' %}`
+{% include 'macro_output_format_narrative' %}
+{% include 'macro_state_emission' %}`
```
failerko: We should exclude this section when we know per-turn classifier will fire. Right now if piggyback set to off in story settings or model isn't set as capable it will generate the block anyway, and then runs per-turn classifier on top of that.

lib/pipeline/definitions/per-turn.ts
```diff
@@ -124,12 +131,54 @@ async function* narrativePhase(ctx: PhaseContext): AsyncGenerator<PhaseEmittedEv
...
-  // inherited values (see submit-turn.ts). M2 has no classifier, so this
-  // propagates the opening's values forward until piggyback/classifier (M3+)
-  // emits fresh ones.
-  const worldTime = tail?.metadata?.worldTime ?? 0
+  const inherited = inheritedEntryMetadata(tail?.metadata)
+
+  const parsedState = parseStateBlock(content)
+  // Fields inside parsedState.block still carry the model's bracketed-ID
+  // placeholders (c1, l1, i1...); swap them back to real entity ids using the
+  // same idMap the prompt was built with before anything looks them up.
```
failerko, line 167: This is an unverified assumption. The substitution system is unable to handle these bracketed IDs and they fall through and eventually leak into the DB and story entry metadata.

So either we establish a convention that ID's are always bracketed as `[c1]` and so on, or we instruct the classifier to output the ID's without brackets.

Second thing this should not fall through, the classifier as designed only ever references existing values, therefore if we don't match a reference to an existing entity it should be filtered out and recorded in `failures`.

failerko (changes requested):

components/compounds/delta-log-row.tsx
Pento95: Fair point, but ai_classifier is the generic "AI-produced" tag used everywhere since M1/M2 (translations, entries, happenings...), not piggyback-specific — that's why they looked like they overlapped. I added a dedicated value, piggyback_tagged_block, just for the direct tagged-block path; ai_classifier stays as-is for everything else.

lib/prompts/bundled/per-turn.ts
Pento95: I didn't notice this when testing, nice catch
Fixed: the narrative phase now resolves whether the fallback will fire BEFORE rendering the prompt (early model resolution, not derived from call.providerId after the stream), and skips staged/locations/calendar/state-emission entirely when it will.

lib/pipeline/definitions/per-turn.ts
Pento95: Claude: That was a real bug, not just a risky assumption: the prompt literally told the model to write "the bracketed ID," so any compliant emission would land outside the expected format and slip through into the DB unnoticed. Fixed the prompt wording (IDs without brackets) and made the substitution layer fail-closed: any unresolved reference — bracketed or not — now gets dropped and logged instead of propagating.

lib/pipeline/definitions/per-turn-piggyback.ts
```diff
@@ -0,0 +1,227 @@
...
+      error: { kind: 'orchestrator', detail: 'piggyback-fallback: no open story' },
+    }
+
+  const outcome = ctx.intermediates.piggybackOutcome as PiggybackOutcome | undefined
+  if (!shouldFallbackFire(outcome)) return { status: 'completed' }
+
+  const entries = [...entriesStore.getEntries().values()]
+    .filter((e) => e.branchId === ctx.branchId)
+    .sort((a, b) => a.position - b.position)
+  const tail = entries.at(-1)
```
failerko, line 122: This is only the very last AI entry, correct? We should also add the last user entry to that, otherwise user actions that should have impact on the world state would do nothing.

lib/pipeline/definitions/per-turn-piggyback.ts
```diff
@@ -0,0 +1,227 @@
...
+  if (!tail) return { status: 'completed' }
+
+  const entities = [...entitiesStore.getEntities().values()].filter(
+    (e) => e.branchId === ctx.branchId,
+  )
+
+  // Same context builder + template pattern as the narrative phase
+  // (lib/pipeline/definitions/per-turn.ts) — the classifier is a
+  // story-related prompt like any other, not a special case.
+  const idMap = new IdBiMap()
```
failerko: Why a new IdBiMap? The per-turn pipeline already initializes it and stores it in intermediates.
This is not a new pipeline, it's just a phase that runs inside that pipeline under some conditions, it can and should re-use the existing map.

lib/prompts/bundled/per-turn.ts
```diff
@@ -22,18 +22,48 @@ export const PER_TURN_NARRATIVE = `{% if definition.setting != blank -%}
...
 {% for e in entities | active -%}
 {%- if sceneEntities contains e.id %}
-## {{ e.name }}
+## {{ e.name }}{% if piggybackFires %} [{{ e.id }}]{% endif %}
 {{ e.description }}
 {% endif -%}
 {%- endfor %}
 
 {% endif -%}
+{%- if piggybackFires -%}
```
failerko: I don't feel like this exclusion is the correct choice, how's the per-turn classifier gonna introduce a new character narratively? 
The instructions on using IDs in the trailing block for sure should be excluded if no piggyback, but the info about locations and staged characters should probably stay, the calendar vocab probably as well so the narrative model can understand the calendar when a date/time is referenced in narrative by the user and so it can the calendar as well in generated narrative.

failerko (changes requested):

failerko (approved):

PR #417, opened 2026-08-06
Image resolution as an intent, custom model entry, and provider option fixes

Pento95:
Split out of #415.

## Resolution is an intent, not a pixel count

The backends disagree on what a size even *is*. Google takes an aspect ratio from a closed
list and has no pixel parameter at all; OpenRouter takes ratio + `1K`/`2K`/`4K`; NanoGPT
publishes a per-model list of accepted values; only ComfyUI, A1111 and Pollinations take
arbitrary dimensions. There is no pixel size all of them accept, so the old pixel selector
was offering the user something most providers could not honour.

It is now an **orientation** (1:1 / 16:9 / 9:16) plus one of four **size steps**, and each
provider adapter resolves that into whatever its backend takes (`src/lib/utils/image.ts` +
`image/providers/*`). `registry.test.ts` covers all nine adapters.

- `specToAspectRatio` falls back to the closest ratio **by shape**, compared in log space —
  on a linear scale portrait ratios crowd into (0, 1] while landscape ones spread over
  [1, ∞), which biases every near-miss towards square. It also seeds its fallback with the
  canonical ratio, so a wholly unparseable `supported` list cannot echo its own garbage back.
- `ImageResolutionSelector`'s accessible name carries the tier and the cost hint and repeats
  the visible dimensions verbatim (WCAG 2.5.3).
- **Stored settings migrate.** `parseImageSpec` reads both the old `WIDTHxHEIGHT` strings and
  the new specs, and `migrateImageGeneration` is idempotent — it runs on every load, not just
  the first after upgrade. Defaults land on the same pixels they replace.

## Model selection

- Manual custom image-model entry, with clearer descriptions and errors. `Autocomplete` now
  says "Type custom value to select..." when it accepts one, instead of "No items found."
- **Hardcoded fallback model lists removed.** A failed fetch now shows an empty list and the
  error, rather than presenting stale model ids as if they were live — which is worse than
  showing nothing, because the user picks one and the request fails later with a less useful
  message.

## `reasoning_effort` never reached any `createOpenAICompatible` provider

The SDK spreads unknown provider options into the request body and *then* assigns
`reasoning_effort` from its own parsed `reasoningEffort`. The later key wins, so the
snake_case value we sent was overwritten with `undefined` and dropped — silently, on every
call, for every provider built that way.

Now sent as `reasoningEffort` for nanogpt, llamacpp, lmstudio, ollama, chutes, nvidia-nim,
openai-compatible and zhipu. Zhipu keeps `thinking`, which is not a key the SDK knows and so
passes through the spread untouched.

Separately, `nvidia-nim` provider options are looked up under `nvidiaNim`; the SDK still reads
the hyphenated form but warns that it is deprecated.

## Second commit — review feedback

- **`InlineImageTracker` used the primary size for reference images.** It switched profile
  *and* model to the reference profile for an img2img generation but kept
  `imageSettings.size`. A reference model is a different model on a different backend, so it
  was handed a size that backend may not accept. `InlineImageService` and
  `aiService.generateImage` already used `referenceSize` — the inline path was the one
  disagreeing. Predates this branch; fixed here because this is the branch that gives `size`
  a type where the mismatch matters.
- `parseImageSpec` takes `unknown`. Its caller is the settings load path, where the value is
  whatever JSON is on disk; a number or boolean reached `value.trim()` and threw — the one
  outcome that path cannot afford.
- `registry.test.ts`'s `ALL_PROVIDERS` claimed to be exhaustive and was not: an array
  annotated `ImageProviderType[]` accepts a short list happily. Now a
  `Record<ImageProviderType, true>`, so the claim is enforced by `svelte-check`.
- Tests for the Zhipu branch of `buildProviderOptions` — the one provider the
  `createOpenAICompatible` loop cannot cover — in both the enabled and `none` cases.

---

Base: `upstream/master`. Conflicts with #418 on `settings.svelte.ts` and `settingsMigrations.ts`
(both add a migration next to the other's) — whichever merges second resolves a three-line
import list. `check`, `lint` and the full Vitest suite pass.


Pento95:
Review triage — thanks. Four taken, three not, with reasons.

**Applied** (`ca78552e`):

- `InlineImageTracker` reference size. Correct and worth more than "nitpick" — it predates this branch, but this is the branch that gives `size` a type where the mismatch matters, so it belongs here.
- `parseImageSpec` scalar guard, with regression coverage for numbers, booleans and arrays.
- `ALL_PROVIDERS` as `Record<ImageProviderType, true>`. Good catch: the existing comment already *claimed* exhaustiveness, and an array annotated `ImageProviderType[]` never enforced it. Now it does.
- Zhipu cases in `generate.test.ts`, both enabled and `none`.

**Not applied:**

- **`openai.ts` / `zhipu.ts` size mapping.** The constraints cited are out of date for the models these adapters target. Separately, the suggested OpenAI list (`1024x1024`, `1536x1024`, `1024x1536`, `auto`) is `gpt-image-1`'s specifically — `dall-e-3` and `dall-e-2` take different sets — so it would not have been correct as a constant even if the premise held.
- **`registry.ts` — don't await `listImageModels` before generating.** That await is the mechanism: NanoGPT resolves the spec against the model's own published resolution list, and the result is served from the same TTL cache the settings UI fills, with `.catch(() => [])` already in place. Reading only a synchronous cache would silently degrade NanoGPT to canonical pixels on the first generation of every session — trading a correctness property for a latency paid once per TTL. If that latency ever matters, the fix is to warm the cache at startup, not to drop the lookup.

Also `aria-live` on the model-refresh error was left for a follow-up.

failerko (approved):

PR #418, opened 2026-08-06
Retrieval: word-budget tiers, retrieval snapshots, and a correct Tier 3 selection cache

Pento95:
Split out of #415. The largest of the four and the one worth the most review time.

Both selection services — **Entry Retrieval** (authored lorebook `Entry[]`) and **World State
Injection** (live tracked entities) — share the same three-tier shape. This reworks the
boundary between "include the leftover whole" and "ask a model which of it matters", and adds
a diagnostic for what actually reached the prompt.

## Include-All Budget

The boundary is now a **word budget on the candidate text**, in both services, replacing World
State's record count (`llmThreshold`).

30 records and 500 words describe the same line only because a live world-state record averages
~16 words. A lorebook entry averages ~69 — which is exactly why the two budgets differ
(500 / 1000) where one shared record count could not express the difference.

- The stored `llmThreshold` is **dropped, not converted**: someone who raised it to 100 was not
  asking for 1600 words of anything, and the new default is calibrated to do what the old
  default did. Dropping the key matters beyond tidiness — the store merges disk over defaults,
  so an unread key is written back into every future save forever.
- Advanced Settings relabels the whole group in those terms: *Include-All Budget*, *Ask the
  Model Above the Budget*, *Max LLM-Selected …*, and stops claiming Tier 3's cap means "there is
  no Tier 3" when selection is off — a leftover under the budget still goes in whole.

## Tier 3 selection

- **The cache returned a different entry set than the one it cached.** It is now keyed on the
  caller, the candidate pool *in order*, and the player's action, and bounded to
  `TIER3_SELECTION_CACHE_POSITIONS` story positions. Order is part of the key because the answer
  lives in index space, and `WorldStateInjector` builds its candidates from world state the
  classifier rewrites every turn. Cleared when the story it was about is no longer loaded, and
  it holds one entry per caller rather than one per question — see the second commit.
- **The model is asked for indices, not ids.** Ids are never rendered into the prompt, so it had
  never seen one. `selectedIds` → `selectedIndices` across the contract, and the prompt states
  the format. `resolveTier3Selection` tolerates `"1."`, `"#1"` and `"[1]"` coming back.
- **Wholesale-included entries no longer count as relevance activations.** They were not chosen
  for relevance, and treating them as if they were kept them sticky afterwards.
- A Tier 2 **second pass** (`tier2SecondPass.ts`) matches what is left against the *names* of
  what the first pass found, so an entry nobody named directly comes in when something that
  names it did. Names and aliases only — descriptions cross-reference each other by
  construction, so seeding with them pulls in half a dense lorebook in one step. Seeds shorter
  than 4 characters, and seeds contained in a longer seed, are dropped.
- `recentContent(..., withRoles)` labels the scene `[Player Action]` / `[Narrator]` for the
  selection prompt, instead of one undifferentiated block. `retry` is labelled `[Narrator]`: a
  model told it was a retry treats the re-roll as an event.

## Retrieval snapshots

`retrievalSnapshot.ts` records what each tier actually contributed, with token counts, and the
narration entry carries it in `metadata.retrievalSnapshot`. The in-memory copy dies with the
session, so without it the debug panel was empty on every fresh start and after every story
switch — exactly when you most want to know what the narrator is being told. Diagnostic only;
nothing reads it back into a prompt.

- The debug panel shows the injected **World State** block alongside the lorebook one, both
  collapsible.
- The toolbar badge counts the whole active context rather than lorebook entries alone, and the
  button is no longer gated on the lorebook having entries — every story has live world state.
  Reachable from the menu on narrow screens.
- **Stickiness is tracked in story positions, not turns.** A turn appends both an action and a
  narration, so a duration of N covers roughly N/2 turns; the panel converts for display.

## Fixes along the way

- `ActionInput` passed the **raw player text** to `generateResponse` while the entry the narrator
  reads holds `promptContent`. With translation on, retrieval and classification worked from a
  different wording than the narration — and the retry path already passed `promptContent`, so
  the two disagreed with each other.
- **Retry re-ran retrieval from scratch** — an entire agentic loop, now that agentic is the
  default mode — because the cached result was cleared before the call site read it.
- `create_entry` / `update_entry` lose `injectionMode` in the **lore management** toolset only.
  It runs unattended, and `always` is a budget decision past every retrieval threshold, made by
  an agent that cannot see the budget. The vault assistant keeps it, where the user reads the
  change before it lands. Removed from the schema rather than ignored afterwards: the schema is
  what the model is shown, so a parameter left in and discarded is a parameter it spends tokens
  choosing.
- The lore management prompt asks explicitly for duplicate consolidation via `update_entry` +
  `delete_entry`, rather than "merge" with no mechanism named.
- `buildWorldStateContext` / `getRelevantLorebookEntries` take an options object; the positional
  `signal` / `activationTracker` pair had already been passed in the wrong order once.

## Second commit — review fixes

- The selection cache held **one entry per distinct question, retained until the story was
  switched**. Only the most recent answer is ever reusable (the window is two positions wide), so
  every earlier key was growth with no hit rate behind it — and each key is every candidate id
  concatenated, several KB on a large lorebook, held for the whole session on a WebView heap that
  is a hard cap on Android. Now one entry per caller, with the question compared against the
  stored key.
- Comments still describing the World State threshold as a record count named `llmThreshold`.
- `splitTier1` and `splitTier3` had their doc comments stacked above `splitTier2`.

### Third commit

`splitTier3` treated a missing `llmSelected` as *selected*, pinned by a test describing "a
snapshot written before the flag existed" — a state that never existed, since the snapshot and
the flag ship in the same commit. All four producers set the flag, so this only decides how a
row from some later source reads; it now reads as wholesale. The two errors are not symmetric:
"LLM Selected" asserts a call was paid for, "Included Whole" only that the leftover fitted.

---

Base: `upstream/master`. Conflicts with #417 on `settings.svelte.ts` and `settingsMigrations.ts`
(both add a migration next to the other's) — whichever merges second resolves a three-line import
list. `check`, `lint` and the full Vitest suite (747 tests) pass.

## Summary by CodeRabbit

- **New Features**
  - Added a combined Active Context view for lorebook and world-state retrieval, including counts, token totals, tiers, carry-over status, reasons, and prompt previews.
  - Added Active Context access on mobile and for all current stories.
  - Added navigation, activation clearing, and demotion controls for retrieved entries.
  - Added configurable recent-entry windows and word budgets for retrieval.

- **Improvements**
  - Improved scene-aware matching, retrieval caching, stickiness, and regeneration consistency.
  - Added clearer retrieval settings descriptions and automatic migration to the new budget controls.
  - Improved lore-management guidance for detecting and merging duplicate entries.

### DIFF
--- a/src/lib/components/debug/LorebookDebugPanel.svelte
+++ b/src/lib/components/debug/LorebookDebugPanel.svelte
@@ -1,27 +1,194 @@
 <script lang="ts">
   import { ui } from '$lib/stores/ui.svelte'
-  import { BookOpen, Users, MapPin, Package, Shield, Lightbulb, Calendar } from '@lucide/svelte'
+  import { story } from '$lib/stores/story.svelte'
+  import {
+    toRetrievalSnapshot,
+    positionsToTurns,
+    splitTier1,
+    splitTier2,
+    splitTier3,
+    type RetrievalSnapshot,
+    type RetrievalSnapshotEntry,
+  } from '$lib/services/ai/retrieval'
+  import { Button } from '$lib/components/ui/button'
+  import {
+    BookOpen,
+    Globe,
+    Users,
+    MapPin,
+    Package,
+    Shield,
+    Lightbulb,
+    Calendar,
+    SquarePen,
+    PanelRight,
+    PinOff,
+    X,
+    ChevronDown,
+  } from '@lucide/svelte'
 
   import * as ResponsiveModal from '$lib/components/ui/responsive-modal'
+  import * as Collapsible from '$lib/components/ui/collapsible'
   import { ScrollArea } from '$lib/components/ui/scroll-area'
   import { Badge } from '$lib/components/ui/badge'
   import { Card, CardContent } from '$lib/components/ui/card'
   import { Separator } from '$lib/components/ui/separator'
   import { cn } from '$lib/utils/cn'
+  import type { SidebarTab } from '$lib/types'
+  import { countTokens } from '$lib/services/tokenizer'
 
-  const tierLabels: Record<number, string> = {
-    1: 'Always Active',
-    2: 'Keyword Matched',
-    3: 'LLM Selected',
+  /**
+   * Live for the turn that just ran, persisted for every other case.
+   *
+   * The in-memory copy belongs to this session, so on a fresh start — or after a story
+   * switch — it is the snapshot on the last narration that has the answer. The live branch
+   * counts its own tokens here, from blocks it still holds; the persisted one carries the
+   * count taken when those blocks were built, since they are not stored.
+   */
+  const snapshot = $derived<RetrievalSnapshot | null>(
+    toRetrievalSnapshot(ui.lastLorebookRetrieval, ui.lastWorldStateRetrieval, countTokens) ??
+      story.entries.findLast((e) => e.type === 'narration')?.metadata?.retrievalSnapshot ??
+      null,
+  )
+
+  const sections = $derived(
+    snapshot
+      ? [
+          {
+            title: 'Lorebook',
+            icon: BookOpen,
+            entries: snapshot.lorebook,
+            tokens: snapshot.tokens?.lorebook ?? 0,
+          },
+          {
+            title: 'World State',
+            icon: Globe,
+            entries: snapshot.worldState,
+            tokens: snapshot.tokens?.worldState ?? 0,
+          },
+        ].filter((s) => s.entries.length > 0)
+      : [],
+  )
+
+  const total = $derived(sections.reduce((sum, s) => sum + s.entries.length, 0))
+  const totalTokens = $derived(sections.reduce((sum, s) => sum + s.tokens, 0))
+
+  /**
+   * Tier 1 is three different things wearing one label, so it is shown as three.
+   *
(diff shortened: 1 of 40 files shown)


src/lib/components/debug/LorebookDebugPanel.svelte
```diff
@@ -1,27 +1,194 @@
...
+      description: "Where you are, who's present, what you're carrying, active quests",
+    },
+    matched: { label: 'Keyword Matched', description: 'Named directly by the scene' },
+    viaScene: {
+      label: 'Matched Through the Scene',
+      description: 'Named by something the scene already pulled in, not by the scene itself',
+    },
+    wholesale: {
+      label: 'Included Whole',
+      description: 'The leftover fitted the budget, so all of it went in — no model asked',
```
failerko: Is `fitted` a correct english for this? 🤔  I'd say "The leftover fit the budget" is more correct.

failerko (approved):

src/lib/components/debug/LorebookDebugPanel.svelte
Pento95: You're right, "fit" is the correct past tense here. Changed it to "fit within budget" so it mirrors "was over budget" on the line below.

PR #419, opened 2026-08-07
fix(import): stop losing entry reasoning, action choices and world-state delta

collegatore:
## Summary

Imported stories lost three `StoryEntry` fields, and a second, older bug kept them from
reaching the screen even once they were preserved. Two commits, one root cause each.

## `fix(import)`: stop dropping entry reasoning, choices and world-state delta

An imported story regenerated its action choices every time it was opened and every time the
user switched branches — a model call each, to reproduce choices the `.avt` file already
contained. Imported entries also rendered no chain-of-thought, and world-state rollback
skipped straight past them.

One cause behind all three: `importStructure` builds its `addStoryEntry` argument field by
field, and `reasoning`, `suggestedActions` and `worldStateDelta` were missing from that object
literal. Nothing else was wrong — the format carries the fields, the INSERT has the columns,
`mapStoryEntry` reads them back. The literal is typed `Omit<StoryEntry, 'createdAt'>` and all
three fields are optional, so omitting them type-checked cleanly and the loss left no trace at
the call site.

Device-to-device sync is fixed by the same change: it imports through the same `runImport`
path, so a synced story no longer arrives stripped.

## `fix(story)`: show persisted action choices when a story is opened

Found while verifying the above. Opening a story whose choices were generated somewhere else
showed no choices at all, and did not regenerate any either — switching branches and back was
the only way to make them appear.

`switchBranch` restores choices from the landed-on narration entry's `suggestedActions`.
`loadStory` never did: it read only the settings-keyed cache this app writes when it generates
choices itself, and that cache has nothing under a story id created moments earlier by an
import. The regeneration fallback did not cover the gap either — `suggestionsRegenerationNeeded`
is set only by `restoreSuggestedActionsAfterDelete`, which `loadStory` never called. So the
choices sat in the database, fully readable, with nothing asking for them.

`loadStory` now falls back to that same restore path when the cache comes up empty, matching
what `switchBranch` already does. The cache still wins when populated, so the common path is
unchanged.

This asymmetry predates the import bug and was merely masked by it, since imported entries had
no choices to find in the first place. Any story whose settings cache is missing while its
entries carry choices would have hit it.

## Notes for review

- **No format-version bump and no migration.** The file already carries these fields and the
  columns already exist, so this is a read-side bug rather than a format extension. Bumping
  would falsely imply older files lack data they may well contain.
- **Absent stays absent**, so regeneration still fires for stories that genuinely have no
  persisted choices. Worth knowing for anyone touching that literal next: `reasoning` is
  declared `reasoning?: string` — optional with no `| null` — while the other two are nullable,
  so coalescing all three to `null` fails `svelte-check`. `addStoryEntry` writes a NULL column
  either way.
- **`worldStateDelta` is passed parsed** — `addStoryEntry` stringifies it, so pre-stringifying
  would double-encode.
- **Scope.** The five sibling entity loops in `importStructure` (characters, locations, items,
  beats, lorebook entries) were deliberately not audited for the same class of omission, to
  keep this reviewable as a targeted fix. Might be worth its own pass.
- Applies to **new** imports. Stories imported before this can be re-imported from their
  original `.avt` to recover the fields; import is non-destructive, creating a new story.

## Testing

- `npm test` — 55 files, 688 tests passing, including 3 new cases: field pass-through,
  absent fields, and entries on a non-main branch.
- `npm run check` — 5501 files, 0 errors.
- Manually verified in a debug desktop build: choices display on open and on branch switch with
  no regeneration triggered, imported reasoning renders, and rollback across imported entries
  applies their deltas instead of reporting them as delta-less.

## Summary by CodeRabbit

* **Bug Fixes**
  * Imported story entries now retain reasoning, suggested actions, and world-state updates.
  * References in imported world-state updates and checkpoint snapshots are remapped correctly, including nested entities and inventory data.
  * Missing optional import fields are handled consistently without causing failures.
  * Reopened stories restore available action choices and suggestions from the latest narration.
  * Invalid, empty, or malformed saved choices and suggestions are safely discarded.

### DIFF
--- a/src/lib/services/import/import.test.ts
+++ b/src/lib/services/import/import.test.ts
@@ -58,7 +58,18 @@ function sampleExport() {
     version: '1.8.0',
     exportedAt: 1,
     story: { id: 'story-old', title: 'T', settings: {} },
-    entries: [{ id: 'entry-1', type: 'narration', content: 'c', parentId: null, position: 0 }],
+    entries: [
+      {
+        id: 'entry-1',
+        type: 'narration',
+        content: 'c',
+        parentId: null,
+        position: 0,
+        reasoning: 'because the hero hesitated',
+        suggestedActions: '["Fight","Flee"]',
+        worldStateDelta: { mood: 'tense' },
+      },
+    ],
     embeddedImages: [
       { id: 'img-1', entryId: 'entry-1', imageData: 'BASE64', prompt: 'p', status: 'completed' },
       { id: 'img-orphan', entryId: 'gone', imageData: 'BASE64', prompt: 'p', status: 'completed' },
@@ -148,6 +159,71 @@ describe('runImport — rollback', () => {
 
     expect(result.success).toBe(false)
     expect(result.error).toContain('the real cause')
+  })
+})
+
+describe('runImport — entry fields', () => {
+  it('carries reasoning, suggestedActions and worldStateDelta through to addStoryEntry', async () => {
+    const result = await runImport(sampleExport())
+
+    expect(result.success).toBe(true)
+    expect(calls.entries[0].reasoning).toBe('because the hero hesitated')
+    expect(calls.entries[0].suggestedActions).toBe('["Fight","Flee"]')
+    expect(calls.entries[0].worldStateDelta).toEqual({ mood: 'tense' })
+  })
+
+  it('imports an entry lacking all three fields without inventing values', async () => {
+    const data = {
+      ...sampleExport(),
+      entries: [{ id: 'entry-1', type: 'narration', content: 'c', parentId: null, position: 0 }],
+    }
+
+    const result = await runImport(data)
+
+    expect(result.success).toBe(true)
+    // `reasoning` is optional-undefined on StoryEntry; the other two are nullable. Either way
+    // `addStoryEntry` writes NULL, and regeneration still fires for a story with no choices.
+    expect(calls.entries[0].reasoning).toBeUndefined()
+    expect(calls.entries[0].suggestedActions).toBeNull()
+    expect(calls.entries[0].worldStateDelta).toBeNull()
+  })
+
+  it('preserves the fields on entries belonging to a non-main branch', async () => {
+    const data = {
+      ...sampleExport(),
+      branches: [
+        {
+          id: 'branch-1',
+          storyId: 'story-old',
+          name: 'side-quest',
+          parentBranchId: null,
+          forkEntryId: null,
+          checkpointId: null,
+          createdAt: 1,
+        },
+      ],
+      entries: [
+        {
+          id: 'entry-1',
+          type: 'narration',
+          content: 'c',
+          parentId: null,
+          position: 0,
+          branchId: 'branch-1',
+          reasoning: 'branch reasoning',
+          suggestedActions: '["Explore"]',
+          worldStateDelta: { location: 'cave' },
+        },
+      ],
+    }
+
+    const result = await runImport(data)
+
+    expect(result.success).toBe(true)
+    expect(calls.entries[0].branchId).not.toBeNull()
+    expect(calls.entries[0].reasoning).toBe('branch reasoning')
+    expect(calls.entries[0].suggestedActions).toBe('["Explore"]')
+    expect(calls.entries[0].worldStateDelta).toEqual({ location: 'cave' })
   })
 })
 
--- a/src/lib/services/import/import.test.ts
+++ b/src/lib/services/import/import.test.ts
@@ -16,6 +16,10 @@ const calls = {
   entries: [] as any[],
   images: [] as any[],
   deleted: [] as string[],
+  characters: [] as any[],
+  locations: [] as any[],
+  items: [] as any[],
+  storyBeats: [] as any[],
 }
 let failOnCreateStory = false
 
@@ -29,10 +33,10 @@ vi.mock('$lib/services/database', () => ({
     createEmbeddedImage: vi.fn(async (i: any) => void calls.images.push(i)),
     deleteStory: vi.fn(async (id: string) => void calls.deleted.push(id)),
     addBranch: vi.fn(async () => {}),
-    addCharacter: vi.fn(async () => {}),
-    addLocation: vi.fn(async () => {}),
-    addItem: vi.fn(async () => {}),
-    addStoryBeat: vi.fn(async () => {}),
+    addCharacter: vi.fn(async (c: any) => void calls.characters.push(c)),
+    addLocation: vi.fn(async (l: any) => void calls.locations.push(l)),
+    addItem: vi.fn(async (i: any) => void calls.items.push(i)),
+    addStoryBeat: vi.fn(async (b: any) => void calls.storyBeats.push(b)),
     addEntry: vi.fn(async () => {}),
     addChapter: vi.fn(async () => {}),
     createCheckpoint: vi.fn(async () => {}),
@@ -49,6 +53,10 @@ beforeEach(() => {
   calls.entries.length = 0
   calls.images.length = 0
   calls.deleted.length = 0
+  calls.characters.length = 0
+  calls.locations.length = 0
+  calls.items.length = 0
+  calls.storyBeats.length = 0
   failOnCreateStory = false
   invoke.mockReset()
 })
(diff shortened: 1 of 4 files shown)


src/lib/services/import/structure.ts
```diff
@@ -135,6 +135,11 @@ export async function importStructure(
       translatedContent: entry.translatedContent ?? null,
       translationLanguage: entry.translationLanguage ?? null,
       originalInput: entry.originalInput ?? null,
+      // `reasoning` is optional-undefined rather than nullable, unlike the two below;
+      // `addStoryEntry` turns the undefined into a NULL column either way.
+      reasoning: entry.reasoning,
+      suggestedActions: entry.suggestedActions ?? null,
+      worldStateDelta: entry.worldStateDelta ?? null,
```
Pento95: `buildIdMaps` mints a fresh UUID for every entity, so every id in the delta matches nothing here. That's not just a rollback that no-ops: `previousState.currentLocationId` reaches `setCurrentLocation`, which runs `UPDATE locations SET current = 0 WHERE story_id = ?` before failing to match the stale id. A regenerate over an imported entry therefore leaves the story with **no** current location, and `WorldStateInjector` stops emitting `[CURRENT LOCATION]`. Silently, since `RollbackSummary` counts a success per attempted row.

The checkpoint block below already states the rule — "whole entity graphs, each needing the same remapping as the live rows above" — and a delta is the same kind of graph. Pushed `remapWorldStateDelta` built on the existing `remapEntityId`, including `items[].location` with its `'inventory'` sentinel; `classificationResult` left alone as debug-only. Same call added to `remapStoryEntry`, or a checkpoint restore hands the stale deltas back.

src/lib/stores/story.svelte.ts
```diff
@@ -616,6 +616,15 @@ class StoryStore {
       await ui.loadSuggestions(storyId)
     }
 
+    // The settings-keyed cache above is empty for a story that never generated choices in this
+    // app (e.g. a fresh import): fall back to the last narration entry's own suggestedActions,
+    // same as switchBranch does, so an imported story doesn't open with choices silently missing.
+    const hasPersistedChoices =
+      story.mode === 'adventure' ? ui.actionChoices.length > 0 : ui.suggestions.length > 0
+    if (!hasPersistedChoices) {
+      this.restoreSuggestedActionsAfterDelete()
```
Pento95: This isn't a getter — on its failure path it also calls `ui.clearActionChoices(storyId)` and sets `ui.suggestionsRegenerationNeeded`, which an `$effect` in `ActionInput.svelte` turns into a model call. So every open of a story without cached choices pays one, creative-writing included even with suggestions off (`refreshSuggestions` doesn't check `disableSuggestions`). A failed or empty generation persists nothing, so it repeats on the next open.

Worst case is a story with no entries yet: the effect's `entries.length > 0` guard holds the flag rather than clearing it, so it fires a second, redundant round of choices right after the first turn's.

Split the pure restore out as `restoreSuggestedActionsFromLastNarration()`; the delete path keeps the clear and the flag.

src/lib/stores/story.svelte.ts
```diff
@@ -616,6 +616,15 @@ class StoryStore {
       await ui.loadSuggestions(storyId)
     }
 
+    // The settings-keyed cache above is empty for a story that never generated choices in this
+    // app (e.g. a fresh import): fall back to the last narration entry's own suggestedActions,
+    // same as switchBranch does, so an imported story doesn't open with choices silently missing.
+    const hasPersistedChoices =
```
Pento95, line 632: Restoring from the last *narration* is right after a delete, but not on open: if the app closed mid-turn the last entry is a `user_action`, and that narration's choices are the ones the player already spent — which then also get written back into the settings cache. Added a guard requiring the last entry to be a narration.

src/lib/services/import/import.test.ts
```diff
@@ -58,7 +58,18 @@ function sampleExport() {
...
+    entries: [
+      {
+        id: 'entry-1',
+        type: 'narration',
+        content: 'c',
+        parentId: null,
+        position: 0,
+        reasoning: 'because the hero hesitated',
+        suggestedActions: '["Fight","Flee"]',
+        worldStateDelta: { mood: 'tense' },
```
Pento95: This is why the delta issue got through: `{ mood: 'tense' }` isn't a `WorldStateDelta` (no `previousState`, no `createdEntities`), the export type is `as any`, and `toEqual({ mood: 'tense' })` passes for any behaviour the import could have. A single `currentLocationId` in it would have failed. Replaced with a realistic delta, plus cases for the remap, the `'inventory'` sentinel and a partial delta.

src/lib/services/import/import.test.ts
```diff
@@ -151,6 +162,71 @@ describe('runImport — rollback', () => {
...
+
+    expect(result.success).toBe(true)
+    // `reasoning` is optional-undefined on StoryEntry; the other two are nullable. Either way
+    // `addStoryEntry` writes NULL, and regeneration still fires for a story with no choices.
+    expect(calls.entries[0].reasoning).toBeUndefined()
+    expect(calls.entries[0].suggestedActions).toBeNull()
+    expect(calls.entries[0].worldStateDelta).toBeNull()
+  })
+
+  it('preserves the fields on entries belonging to a non-main branch', async () => {
```
Pento95, line 281: Nit: this doesn't test anything the first case doesn't — `branchId` has no interaction with the three fields. Repointed it at something branch-specific: that the delta remap isn't skipped for an entry on a branch.

Pento95 (changes requested):
Thanks — good catch. The missing `reasoning` / `suggestedActions` are real: the export does write all three fields, so `importStructure` was the only place they were lost. I also checked the five sibling loops you flagged as unaudited — they all write every field, so that note can be closed.

Requesting changes on two data-flow issues, detailed inline:

1. **`worldStateDelta` can't cross an import unmapped** — it doesn't just fail to roll back, it clears the imported story's current location.
2. **`loadStory` picks up two side effects** from `restoreSuggestedActionsAfterDelete`, one of which spends a model call on every story open.

Off the changed lines: the restored blob reaches the UI as a bare `parsed as ActionChoice[]`, and one choice missing `type` crashes `ActionChoices.svelte` — reachable from an external `.avt` only because of this PR. Fixed with the zod schemas that already exist.

Pre-existing, noted so they aren't lost: `ui.suggestions` is never cleared when opening an adventure story, so it renders the previous story's; the checkpoint snapshots carry the same unmapped delta (fixed here too); `refreshSuggestions` ignores `disableSuggestions`; `RollbackSummary` counts zero-row writes as successes.

I've pushed the fixes to this branch — treat the inline comments as the reasoning behind them, not a to-do list.

Pento95:
Thanks again — the missing `reasoning` / `suggestedActions` were a real bug.

Pushed three commits rather than leaving a to-do list:

- **`worldStateDelta` remap** — the import mints fresh ids, so the delta arrived pointing at nothing; `currentLocationId` reaching `setCurrentLocation` left imported stories with no current location at all.
- **Restore on open without arming regeneration** — the delete-path helper also clears and requests regeneration, costing a model call per open.
- **Validate the blob before restoring** — one choice missing `type` crashes `ActionChoices.svelte`, now reachable from external files.

Fixtures updated too: the old `{ mood: 'tense' }` delta couldn't fail. 691 passing, `check` clean. Reasoning is in the review — happy to rework any of it.

src/lib/stores/ui.svelte.ts
```diff
@@ -1026,18 +1028,32 @@ class UIStore {
...
+      // an imported `.avt` or a synced device, and nothing between the file and here checks its
+      // shape. `ActionChoices.svelte` looks up its icon by `choice.type` and renders the result
+      // unconditionally, so a single entry missing that field takes down the whole view.
       if (storyMode === 'adventure') {
-        this.actionChoices = parsed as ActionChoice[]
+        const validated = z.array(actionChoiceSchema).safeParse(parsed)
+        if (!validated.success) {
+          console.warn('[UI] Discarding malformed saved action choices:', validated.error)
+          return false
+        }
```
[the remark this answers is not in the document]
collegatore: Applied in `9c0e354`, and widened: the same method returns `false` from two further paths the suggestion didn't cover — an empty parsed array, and the `JSON.parse` catch — which skipped the clear as well, so patching only the two `safeParse` branches would have left the inconsistency one branch over. All four now go through a single `clearForMode()`, making the contract uniform: a `false` return leaves nothing set.

One correction to the rationale, though the change stands: no current caller can observe the stale state. `restoreSuggestedActionsAfterDelete` clears explicitly right after a `false`, and the open path in `loadStory` only runs when the cache is already empty, so the time-travel scenario described isn't reachable today. This fixes the invariant rather than live behaviour.

src/lib/services/import/import.test.ts
```diff
@@ -151,6 +197,184 @@ describe('runImport — rollback', () => {
...
+    const data = sampleExport()
+    data.checkpoints = [
+      {
+        id: 'cp-1',
+        storyId: 'story-old',
+        name: 'before the cave',
+        lastEntryId: 'entry-1',
+        lastEntryPreview: 'c',
+        entryCount: 1,
+        entriesSnapshot: [data.entries[0]],
```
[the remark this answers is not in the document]
collegatore: Cloned in `f30bde0`.

Worth recording that the test was not passing for the wrong reason: nothing mutates the source entry today — the live loop builds a fresh object literal and `remapWorldStateDelta` returns a new object — and removing the checkpoint remap alone does fail this case, which is how the test was checked when it was added. The reference sharing was a latent coupling rather than an active hole: it left the guarantee resting on import never mutating its input, so an in-place remap added later would have satisfied the assertion through the live path with the snapshot remap gone. Cloning removes that dependency, and the fail-on-removal check still holds afterwards.

collegatore:
Is anything still required from my side? I see a requested change status, but all comments are documentary

Pento95 (approved):

PR #421, opened 2026-08-07
Branch navigation: configurable branch-switch landing (Labs) and an in-branch fork point marker

collegatore:
Two related additions to how the story view orients the reader around branches.

## 1. Branch switch landing (Labs, off by default)

Switching branches left the story view's scroll position and virtual entry window wherever they happened to be — `StoryView` only resets them when the *story* ID changes, so a branch switch inherited the previous branch's `windowStart`/`windowEnd`. Landing was incidental, and switching from a long branch to a short one could clamp the stale window to an empty slice, leaving a blank view showing only "N earlier entries hidden".

Adds a Labs group **Branch Switch Landing**: a master toggle (default off, so current behaviour is untouched) and a target of **Last entry** or **Branching checkpoint**.

- `ExperimentalFeatures` gains `branchSwitchLanding` and `branchSwitchLandingTarget`; the existing spread-merge on load makes this a no-op for current installs.
- `story.switchBranch` emits a new `BranchSwitched` event once entries and caches are correct — before the background-image and suggested-action restores, so a failure in either can't swallow it. Covers both the normal and `skipReload` paths.
- `StoryView` subscribes and lands. **Last entry** anchors the window to the end and scrolls the latest entry itself into view — not the container bottom, since action choices and padding render below it and would push the entry off screen. **Branching checkpoint** resolves `Branch.forkEntryId`, windows around it and scrolls it near the top, falling back to last-entry for the main branch or a missing fork entry.
- A switch that happens while the story panel is hidden is deferred and consumed when the panel returns.

## 2. Fork point marker

Checkpoints are scoped to the branch that owns them, so a checkpoint made on main shows its bookmark and amber "Branch from here" button on main only. From inside the branch, the entry it diverged at looked like any other entry.

Marks that entry with a `GitBranch` icon in the theme accent, resolved from the active branch's `forkEntryId`. Icon only, branch name in its `title`; nothing is marked on the main branch, which has no `Branch` record.

The marker reproduces the "Branch from here" button's `h-7 w-7` slot so the two line up, but renders just outside the action-toolbar container: that container is guarded on `entry.type !== 'system'` plus the editing flags, and a fork entry can legitimately be a system entry. It's a `span`, not a `Button`, and wears the accent rather than the amber reserved for actionable controls, so it doesn't read as clickable.

Checkpoint scoping, branch creation and checkpoint creation are unchanged.

## Notes

- No database migrations, no Rust/Tauri changes, no new dependencies.
- `npm run check` clean; 688 tests pass.
- Verified manually in a release build: both landing targets, main-branch fallback, persistence across restart, Labs reset, and the marker in a light and a dark theme.

## Summary by CodeRabbit

* **New Features**
  * Added experimental controls for repositioning the story view after switching branches.
  * Choose to land on either the branch’s last entry or its fork point.
  * Added a visual marker showing where the active branch diverged from its parent, with the branch name available in its tooltip.
  * Branch landing preferences are saved and restored automatically, with destination controls disabled when the feature is off.
  * Branch switches now reliably load the selected branch before restoring related content.

### DIFF
--- a/src/lib/components/story/StoryEntry.svelte
+++ b/src/lib/components/story/StoryEntry.svelte
@@ -282,6 +282,15 @@
   )
   const canBranch = $derived(!!entryCheckpoint)
 
+  // Fork point marker: the entry the ACTIVE branch diverged from its parent at.
+  // Independent of entryCheckpoint above — that checkpoint belongs to the parent
+  // branch, so it is deliberately invisible from in here (see Branch.forkEntryId).
+  const activeBranch = $derived(
+    currentBranchId ? story.branches.find((b) => b.id === currentBranchId) : undefined,
+  )
+  // Main branch has no Branch record, so no divergence point
+  const isForkPoint = $derived(!!activeBranch && activeBranch.forkEntryId === entry.id)
+
   // Handle creating a branch from this entry
   async function handleCreateBranch() {
     if (!branchName.trim()) return
@@ -1184,6 +1193,18 @@
 
     <!-- Spacer to push buttons to the right -->
     <div class="flex-1"></div>
+
+    <!-- Fork point marker: this branch diverged from its parent here. Passive (not a
+         Button) and outside the toolbar guard below, so it survives on system entries.
+         Matches the toolbar's h-7 w-7 slot so it lines up with "Branch from here". -->
+    {#if isForkPoint}
+      <span
+        class="flex h-7 w-7 shrink-0 items-center justify-center"
+        title="Branch &quot;{activeBranch?.name}&quot; starts here"
+      >
+        <GitBranch class="text-accent-500 h-4 w-4" />
+      </span>
+    {/if}
 
     <!-- Right side: Action buttons toolbar (always visible on mobile, hover-only on desktop) -->
     {#if !isEditing && !isDeleting && !isBranching && !isCreatingCheckpoint && entry.type !== 'system'}
--- a/src/lib/components/story/StoryEntry.svelte
+++ b/src/lib/components/story/StoryEntry.svelte
@@ -1196,13 +1196,16 @@
 
     <!-- Fork point marker: this branch diverged from its parent here. Passive (not a
          Button) and outside the toolbar guard below, so it survives on system entries.
-         Matches the toolbar's h-7 w-7 slot so it lines up with "Branch from here". -->
+         Matches the toolbar's h-7 w-7 slot so it lines up with "Branch from here".
+         Muted rather than accent-coloured: --color-accent-500 is the exact amber of the
+         "Branch from here" button in some themes (fantasy, royal), and both can land on
+         the same entry when a branch's fork entry is still its latest entry. -->
     {#if isForkPoint}
       <span
         class="flex h-7 w-7 shrink-0 items-center justify-center"
         title="Branch &quot;{activeBranch?.name}&quot; starts here"
       >
-        <GitBranch class="text-accent-500 h-4 w-4" />
+        <GitBranch class="text-muted-foreground h-4 w-4" />
       </span>
     {/if}
 
--- a/src/lib/components/story/StoryView.svelte
+++ b/src/lib/components/story/StoryView.svelte
@@ -11,6 +11,7 @@
   import ActionChoices from './ActionChoices.svelte'
   import { Button } from '$lib/components/ui/button'
   import EmptyState from '$lib/components/ui/empty-state/empty-state.svelte'
+  import { eventBus, type BranchSwitchedEvent } from '$lib/services/events'
 
   const storyMaxWidthStyle = $derived.by(() => {
     const maxWidth =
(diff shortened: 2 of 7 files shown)


src/lib/stores/story.svelte.ts
```diff
@@ -3835,6 +3836,11 @@ class StoryStore {
     this.invalidateWordCountCache()
     this.invalidateChapterCache()
 
+    // Announce as soon as entries are correct (both the reload and skipReload paths).
+    // Emitted before the background/suggestion restores below so a failure in either
+    // can't silently swallow the notification.
+    eventBus.emit<BranchSwitchedEvent>({ type: 'BranchSwitched', branchId })
```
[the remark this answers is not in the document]
collegatore: Fixed in a42babc — `createBranchFromCheckpoint` now calls `switchBranch(branch.id)` so entries reload before `BranchSwitched` is emitted, and the redundant follow-up reload is gone; I also dropped the now-unused `skipReload` parameter so the stale-entry window cannot be reintroduced (this incidentally fixes the suggested-action restore, which ran against the outgoing branch on that path).

src/lib/stores/story.svelte.ts
```diff
@@ -3809,10 +3809,11 @@ class StoryStore {
    * Switch to a different branch.
    * This reloads entries from the database filtered by the target branch.
    * NO data is deleted - branches coexist in the database with different branch_ids.
+   * Entries are always reloaded before BranchSwitched is emitted, so subscribers
+   * never observe the previous branch's entries.
    * @param branchId - The branch to switch to (null for main branch)
-   * @param skipReload - If true, skip reloading entries (used when creating new branch from current state)
    */
-  async switchBranch(branchId: string | null, skipReload: boolean = false): Promise<void> {
+  async switchBranch(branchId: string | null): Promise<void> {
```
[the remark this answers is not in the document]
collegatore: Fixed in 34617af — `switchBranch` now queues each call behind the previous one via a promise chain, so overlapping switches can no longer interleave their reloads or emit `BranchSwitched` against another branch's entries; the chain is kept resolved so a failed switch cannot poison later ones while the caller still observes its own error.

Pento95:
Thanks for this — the fork point marker is a nice idea, and it works well.

I'd like to change the landing part before merging, and I want your take first.

The setting has two fields for three outcomes, and I don't think any of them should be a setting:

- With the toggle off (the default), switching from a long branch to a short one still leaves the view blank — just "N earlier entries hidden". Your PR fixes that, but only when the flag is on, so by default everyone keeps the bug. That fix should always be on.
- "Last entry" is what people want almost every time.
- "Branching checkpoint" as a mode sends you back to the fork on every switch, even when you return to the branch you're writing in — far back, and with auto-scroll off. It feels like a one-time jump, not a setting.

So my proposal: drop both fields, always land on the last entry, and add a button in the Branches side panel to jump to the fork point. The marker stays as it is.

Questions:

1. What was your use case for landing on the fork point? If you have a real workflow for it, tell me — I may be missing something.
2. Did you try the long → short branch switch with the flag off?
3. One thing to fix either way: the marker uses text-accent-500, but --color-accent-500 is #f59e0b in the Fantasy theme — the exact amber of the "Branch from here" button. Same colour, same icon, same row, so there it reads as a disabled button. Could you pick a colour that can't collide?
4. Rapid switches: the queue runs every switch in full. Would you add "last one wins" so only the final switch reloads?

src/lib/stores/story.svelte.ts
```diff
@@ -3805,14 +3805,36 @@ class StoryStore {
...
+    // Overlapping switches would interleave their reloads: reloadEntriesForCurrentBranch
+    // reads currentBranchId when it starts and assigns this.entries when its queries
+    // resolve, so the slower call assigns last and leaves entries out of step with
+    // currentBranchId — with BranchSwitched announcing a branch whose entries aren't
+    // loaded. Queue each switch behind the one before it.
+    const run = this.branchSwitchChain.then(
```
Pento95, line 3834: The queued switch isn't bound to the story it was requested for: `performBranchSwitch` reads `this.currentStory` when it runs. Open another story while a switch is still queued and `switchBranch(null)` writes the branch onto the new story — a non-null id is caught by the validation below, `null` isn't. Capture the story id here and bail if it changed.

src/lib/components/story/StoryView.svelte
```diff
@@ -185,6 +186,100 @@
     })
   }
 
+  // ===== Branch switch landing (Labs: branchSwitchLanding) =====
+
+  // Entries kept above the fork entry so the checkpoint has a little context
+  const FORK_CONTEXT_BEFORE = 5
+
+  // Landing requested while the story panel was hidden; consumed when it returns
+  let pendingBranchLanding: { branchId: string | null } | null = null
```
Pento95, line 212: Not cleared when the story changes, so returning to the story panel lands using a branch id from the previous story.

src/lib/components/story/StoryView.svelte
```diff
@@ -185,6 +186,100 @@
...
+
+  // Scroll a specific entry to the top of the viewport. Falls back to the container
+  // bottom when the element isn't in the DOM (entry outside the virtual window).
+  function scrollEntryIntoView(entryId: string) {
+    suppressScrollHandler = true
```
Pento95: One boolean, reset two frames later. With two landings overlapping, the first reset lands mid-scroll of the second, and that scroll is read as a user scroll — which turns auto-scroll off. A token or counter per call fixes it.

src/lib/components/story/StoryEntry.svelte
```diff
@@ -1185,6 +1194,18 @@
...
 
+    <!-- Fork point marker: this branch diverged from its parent here. Passive (not a
+         Button) and outside the toolbar guard below, so it survives on system entries.
+         Matches the toolbar's h-7 w-7 slot so it lines up with "Branch from here". -->
+    {#if isForkPoint}
+      <span
+        class="flex h-7 w-7 shrink-0 items-center justify-center"
+        title="Branch &quot;{activeBranch?.name}&quot; starts here"
+      >
+        <GitBranch class="text-accent-500 h-4 w-4" />
```
Pento95: `--color-accent-500` is `#f59e0b` in the Fantasy theme — the exact amber of the "Branch from here" button. Same colour, same icon, same row, so there the passive marker reads as a disabled button. In `dark` it's `#3b82f6`, the same blue as "Create checkpoint". Worth a colour that can't collide.

Pento95 (changes requested):
The fork point marker is a good idea and works well. Requesting changes for the landing part — see my comment above for the settings proposal, plus three small bugs inline.

src/lib/components/story/StoryEntry.svelte
collegatore: Initially I thought about something like `green-500`, but opted for the the idea to follow the color theme. I chose `--color-accent-500` because it is used in the branches panel to mark an active branch with a dot. Also `green-500` is taken by check mark on copy (I guess it's fine), text color in `Retro Console`, wich is problematic, and branching button in `Botanical` which is the same issue that `--color-accent-500` has now. Default grey is fine, but it's easy to miss that an entry has more badges than the previous or the next one. I honestly don't have a knack for visual design, so there's no strong opinion from my side of what it should be.

collegatore:
About landing settings: I am partially happy with the outcome, and I wanted someone to take a look before forcing anything as default, so the "Labs" felt right. 
- The "last entry" now lands at the top of the entry, not all the way down to the action choices. It feels more useful for me to rehearse what was happening in the branch if I jump from a highly diverged point, but that might not be the natural preference for everyone. 
- I agree that the "branching point" as an ongoing behavior isn't a great choice (it is confusing even for me sometimes), and making it a manual control fulfills all I wanted from it. However, on an Android phone, the branches tree is pretty cramped already; the branch names past the second level of nesting have barely three symbols to display. Probably a `^` button near the `+` in the header row will do, but I don't know how to explain its effect to a player when there are no tooltips on Android. Would a pop-up message "Jumped to the branch origin" akin to "Copied to clipboard" fit the general design?

For the questions:
1. Sometimes I have stories that have parted, say, on entry 30, and have 200+ entries each past that point, with a completely detached set of events. Finding an entry point helps a lot to regain the context when I switch between them. Now that you asked, I acknowledge that one-time jump fits the scenario better than my approach.
2. Long->short branch jump is a real bug. Switching from 200 messages to 40 produces an empty screen with "load previous 40 messages" on top, which I have to click on several times before I reach the text. I would like to address it; the real question is "all the way down" vs "top of the last message".
3. As per my comment in the discussion branch, my only safe bet is grey. Suboptimal, but safe.
4. I didn't consider the scenario; in my case, the reload was always lightning-fast, but I will take a closer look at this part.

src/lib/stores/story.svelte.ts
collegatore: Fixed in b85bf17 — `switchBranch` now captures the story id at request time and `performBranchSwitch` bails when `currentStory.id` no longer matches, so a queued `switchBranch(null)` can no longer write onto a story opened in the meantime.

src/lib/components/story/StoryView.svelte
collegatore: Fixed in b85bf17 — the story-switch effect now clears `pendingBranchLanding`, so a landing deferred in one story is dropped rather than applied after opening another.

src/lib/components/story/StoryView.svelte
collegatore: Fixed in b85bf17 — each programmatic scroll now takes an owner token via `suppressScrollFor()` and only the newest owner clears the flag, so an earlier landing can no longer unsuppress mid-scroll of a later one; `scrollToTop` uses the same helper.

src/lib/components/story/StoryEntry.svelte
collegatore: Already fixed in a764ada (pushed just before this review) — the marker is now `text-muted-foreground`, which is a desaturated neutral in every theme, so it collides with neither the amber "Branch from here" nor the blue "Create checkpoint". Worth noting your point also holds for a case my design doc wrongly ruled out: while a branch has no new entries its fork entry is still the latest entry, so a checkpoint can be created there and the amber control renders on the same entry as the marker.

Pento95 (approved):

PR #437, opened 2026-08-09
Lore management: a session ledger, a shared chapter-read budget, and a duplicate consolidation window

Pento95:
Two commits: the lore management refactor, and a pass over it.

## Correctness

- **`foldName` folded every Cyrillic, Greek and CJK name to the empty string**, so `sameEntityName` reported any two of them as the same entity, across the classifier's fourteen call sites. Folds on `\p{L}\p{N}` now, like the rest of `text.ts`.
- **A rendered entity line was re-readable as a name.** The classifier was shown `- Eira (claimed as a consort) [inactive]` and returned that whole string as the name; `sameEntityName` missed, and a second character was created — four of thirty-eight on a measured 41-chapter save, two carrying the subject's own `relationship` verbatim. Name and attributes are separate lines now, here and in the narrator block whose prose the classifier also reads.
- **Lore changes go through `LoreSessionLedger`**, one slot per index of the array the tools index into. A create-then-update lands as one create, an update-then-delete as one delete, and a second update to the same entry keeps the first — all silently dropped before, after the tool had already answered `success: true`. A merge carries `hiddenInfo` from every source, which the agent is never shown and so could not carry itself.
- A duplicate group counts as resolved when it has **collapsed to one member**, not when a member was touched: an update is what the agent does anyway, so the obligation opened its own gate.
- `keep_separate` must name every index of a group; closing on one shared index dismissed neighbours the agent had never read.
- `create_entry`'s duplicate refusal matches on `foldName`, not the detector's deliberately lenient `normalizeName`, which read "De Luca" as "Luca" and refused a legitimate entry.

## Duplicate consolidation

`services/duplicates` groups both pools. The world state had no detector at all and is where duplicates actually accumulate — the classifier mints a new `Character` for every new title, so one save held `Baron Kaelen` and `Forge-Master Kaelen`, `Captain Vor'koth`, `General Vor'koth` and `The Captain`.

A **Duplicates** window in the Active Context panel resolves them one group at a time. A merge is shown as a plan before it is written: every field carries where its value came from (`only`, `agreed`, `union`, `conflict`), conflicts are picked by the user, and prose can keep both. Deleting rows is not undoable, so nothing is decided silently — and there is no "the newer row wins", because `characters`, `locations` and `items` have no creation timestamp.

Dismissals persist in `kept_separate` (migration 037), keyed by normalized name pair and scoped to a branch. The lore agent reads the same table and writes its own `keep_separate` decisions to it, so neither side re-argues a settled question.

## Cost and concurrency

- **`ChapterQueryBudget`** is shared by both agents that can read a whole chapter: a budget, a repeat cache, and failures cached like answers. Retrieval spends 3 per turn, lore management 6 per session, and the refusal names `grep_chapters` only where that tool is registered. Removes the duplicate dedup wrapper in `AgenticRetrievalService`.
- **One lore session per branch**, enforced in the coordinator rather than in a UI flag that lingers two seconds after a run. Manual chapter creation is refused while a turn's background tasks are building one.
- The session's inputs are read **when it starts**, not when the turn did — the classifier and the chapter check both run in between, so the snapshot was missing the entries just classified and the chapter that triggered the run.
- `finishOnlyOnLastStep` applies to lore management too, so hitting the ceiling costs the account of the run rather than the run.

## Removals

- The Tier 1 "by state" branch in `EntryRetrievalService`: `isPresent`, `isCurrentLocation`, `inInventory` and faction standing were initialized blank by every path and written by nothing, so none of the four conditions ever fired.
- `mentionCount` / `firstMentioned` / `lastMentioned`, never produced. The columns stay, with their defaults.
- Three unused stop-condition factories, and two `ContextWindowSettings` keys with no reader.
- `recentEntriesForRetrieval` → `recentEntriesForSuggestions`: it named the one thing it does not drive. Migrated for anyone who had tuned it.

Defaults live in `core/defaults.ts` as `*_DEFAULTS`, and consumers never write `?? <default>` — that form had put four stale copies of `maxIterations` in the tree at once.

**+69 tests.** `check`, `lint` and the suite are green; lint warnings are unchanged from the baseline.

## Summary by CodeRabbit

- **New Features**
  - Added duplicate-entity review, merge previews, and “keep separate” decisions for characters, locations, items, and lore entries.
  - Added manual lore-management controls, progress summaries, and safeguards against concurrent runs.
  - Added shared chapter-query limits and caching.

- **Bug Fixes**
  - Prevented manual chapter creation during active background work.
  - Improved name matching, duplicate handling, and lore-field cleanup.
  - Migrated context settings while preserving existing preferences.

- **Documentation**
  - Streamlined the product overview, setup instructions, development guidance, and technology information.


Pento95:
@coderabbitai Checked each finding against current code; fixes are in the follow-up commit.

**Accepted:** empty `pairKeys` — the real one: a group of identically named rows was dropped before it was ever offered, in the window and in the agent's worklist alike. Plus `injectionMode` not mirrored onto the tool-visible entry, unawaited `onKeepSeparate`, `sameEntityName` on an empty fold, `applyMergePlan` writing `null` over the non-nullable `visited`/`status`, `onForget` without a `finally`, `kept_separate.branch_id` nullable inside the PK (SQLite doesn't enforce it, so `INSERT OR IGNORE` appended a row per repeated dismissal), `delete_entry` missing from the duplicate-workflow instructions, `list_entries`' error shape, and the three doc fixes.

**Declined:** `resetAllSettings` never touched `serviceSpecificSettings` at all — pre-existing. `mergeGroup` atomicity: the store methods also drive reactive state, so wrapping them is riskier than the failure. Blocking in `findDuplicateGroups`, in-flight dedup in `chapterQueries` (tools run sequentially), `'empty'` `FieldOrigin`, the redundant `.filter(Boolean)`, comment reordering: cosmetic or premature. On `loreSession`: fair, but the thunk exists to read fresh — `storyId` now comes from the store with the rest of the snapshot instead.

Two it missed: the group key wasn't pool-qualified, so a `Character` group and a lorebook `Entry` group for one person collided in a keyed `{#each}`; and `foldName` split words on the apostrophe instead of joining across it, so `Kaelen's Rest` didn't match `Kaelens Rest` — an example its own doc comment claimed worked.

src/lib/utils/text.ts
```diff
@@ -611,3 +611,49 @@ export function expandRangeBidirectional(
...
+ *
+ * Articles are kept, and that is the whole difference from `normalizeName`. Two names that
+ * differ by an article are the same *subject* but not the same *trigger*: matching is
+ * literal and whole-word, so the alias "The Citadel" fires on that two-word phrase while
+ * the keyword "Citadel" fires on the bare word, and neither makes the other redundant.
+ * Lorebook-entry *identity* is judged by `normalizeName`, which does strip them.
+ *
+ * The character class is `\p{L}\p{N}`, like the rest of this file, and deliberately not
+ * `a-z0-9`: the ASCII form folds every Cyrillic, Greek and CJK name to the empty string,
+ * and empty compares equal to every other one.
```
a-frazier, line 626: Dumb American here: Is this really the best choice?

a-frazier (approved):

src/lib/utils/text.ts
Pento95: wdym?

failerko: Probably fine i think. It seems like lorebooks are completely exempt from translation so it should be eng only, which should be fine for this...

The fact lore management output is exempt from translation is an issue of it's own but oh well.

PR #439, opened 2026-08-10
Sync: multi-select pull/push in the sync modal

MrLinks75:
Adds checkboxes to the Pull/Push story lists so multiple stories can be
synced in one go instead of one at a time. Each story is still pulled or
pushed via its own authenticated request, conflict check, and pre-sync
backup, run sequentially, the batch is just a loop over the existing
per-story flow.

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting and transferring multiple stories at once.
  * Added select-all controls for streamlined batch selection.
  * Added independent pull and push actions for greater control.
  * Added conflict previews before pulling stories.
  * Added per-story progress tracking with success and failure reporting.
  * Added clearer batch transfer status for both pull and push operations.
  * Improved story replacement handling during transfers.

### DIFF
--- a/src/lib/components/sync/SyncModal.svelte
+++ b/src/lib/components/sync/SyncModal.svelte
@@ -139,16 +139,19 @@
     showReceivedConflict = false
 
     try {
-      // If replacing, delete the existing story first
+      // Look up the existing same-titled story before importing: the freshly imported
+      // story shares that title and sorts first by updated_at, so finding it afterward
+      // would return the new story's own id instead of the one to replace.
       const existingId = await syncService.findStoryIdByTitle(receivedStoryPreview.title)
-      if (existingId) {
-        await syncService.createPreSyncBackup(existingId)
-        await syncService.deleteStory(existingId)
-      }
 
+      // Import the replacement before deleting anything, so a failed import never
+      // leaves the device with neither copy.
       const result = await exportService.importFromContent(receivedStoryJson, true)
 
       if (result.success) {
+        if (existingId) {
+          await syncService.deleteStory(existingId)
+        }
         await story.loadAllStories()
         syncSuccess = true
         syncMessage = `Successfully received "${receivedStoryPreview.title}"`
@@ -347,6 +350,9 @@
   }
 
   function toggleRemoteSelection(id: string) {
+    // A pending conflict warning was checked against the selection as it stood a moment
+    // ago; any change here invalidates it, so force a fresh check on the next attempt.
+    cancelConflict()
     const next = new SvelteSet(selectedRemoteIds)
     if (next.has(id)) next.delete(id)
     else next.add(id)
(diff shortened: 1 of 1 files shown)


a-frazier (approved):

a-frazier:
### Code review

Found 1 issue:

1. Batch-pulling two remote stories that share the same title silently destroys the first one pulled. Each loop iteration calls `findStoryIdByTitle(s.title)` (which returns the most-recently-updated story with that title, per `getAllStories()`'s `ORDER BY updated_at DESC`), so on the second iteration the lookup returns the story just imported by the first iteration, and the delete-the-old-copy step removes it — while `succeeded` already recorded it as pulled. The upfront conflict check only compares selected titles against existing local stories, not against duplicate titles within the selection itself, so nothing warns the user. This couldn't happen before the PR since only one story could be pulled at a time.

https://github.com/AventurasTeam/Aventuras/blob/29ef984c9d225a226ec87364c08f3753002e68fa/src/lib/components/sync/SyncModal.svelte#L446-L468

A pre-scan of `storiesToPull` for duplicate titles (surfacing them in the conflict warning), or resolving all `existingId`s before the loop starts, would close the gap.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Pento95:
Thanks for this — import-before-delete genuinely fixes a case where a failed download left the device with neither copy, and restoring the previously open story after a push batch is a good catch. +1 to @a-frazier's duplicate-title finding; a few more things I'd like you to check, in case I've misread the flow.

**The pre-sync backup is gone from both pull paths.** `master` called `createPreSyncBackup(existingId)` before `deleteStory(existingId)` in `pullStory` and `importReceivedStory`; only the push one survives. The new ordering protects against a failed import, but the backup protected against something else — pulling an older copy over newer local work — and that's now unrecoverable. The conflict warning was reworded away from it too, so it isn't visible from the UI. Intended?

**A throw outside the per-item try leaves the modal stuck.** `pullSelectedStories` has `try/finally` but no `catch` (master wrapped the whole body). If `await story.loadAllStories()` throws, `reportBatchResult` never runs and `setSyncMode('connected')` is skipped — and the template renders the spinner for `syncMode === 'syncing'` regardless of `loading`, so it's a permanent spinner with no footer and no Done. Same shape on push, where `story.closeStory()` sits unguarded before `loading = false`.

**A partial failure renders as a full success.** `reportBatchResult` sets `syncSuccess = true` as soon as one story succeeds, and the success block is a hardcoded green check with "Sync Complete!" — 1-of-5 and 5-of-5 look identical, with the failures only in the body text.

**Double-clicking "Pull N Selected" can start two batches.** The conflict pre-check runs N sequential `await checkStoryExists` calls *before* `loading = true`, and the button is only disabled on `!loading` — so it stays live for the whole pre-check, and that window grows with the selection size. Setting `loading = true` first would close it.

Heads-up: this and #446 both rewrite `pullStory` and `importReceivedStory` in the same file, so whichever lands second needs a real merge.

PR #444, opened 2026-08-12
Fix: Story pack template resolution, chapter summary schema length enforcement

### LINKED ISSUE #443: Suggestion agent
<kbd>[Centipede](https://discord.com/channels/1456625039846805686/1537122551682302002/1537122551682302002)</kbd> [Centipede](https://discord.com/channels/1456625039846805686/1537122551682302002/1537122551682302002)  `BOT`

So, my suggestions prompts seems not to be working when applied to a custom prompt pack, it only works if I change the default prompt pack

---
[View on Discord](https://discord.com/channels/1456625039846805686/1537122551682302002/1537122551682302002)

failerko:
## Summary by CodeRabbit

* **New Features**
  * AI-generated content, translations, retrieval, and image prompts now use the selected story or pack context.
  * Added consistent style-prompt resolution with fallback support for custom and default packs.
  * Wizard generation and character-card tools now preserve the selected pack context.

* **Bug Fixes**
  * Prevented story switches during generation from redirecting in-progress AI tasks.
  * Improved consistency and accuracy of story-specific prompts across workflows.
  * Updated chapter summaries to follow configured prompt verbosity guidance.

### DIFF
--- a/docs/architecture/prompts.md
+++ b/docs/architecture/prompts.md
@@ -19,6 +19,34 @@ stored under the id `<template-id>-user`, and `ContextBuilder.render(id)` return
 resolving both. A service that destructures only `system` silently drops the user half — every service
 except the two whose templates deliberately have none.
 
+### Which pack is "the active pack"
+
+There is no ambient one. A pack is chosen per story (`stories.pack_id`), and the wizard holds its own
+selection until the story it is building exists. A `ContextBuilder` therefore has to be told, and there
+are three ways to build one:
+
+| Factory                          | Use when                                                             |
+| -------------------------------- | -------------------------------------------------------------------- |
+| `ContextBuilder.forStory(id)`    | the template also wants story variables (mode, pov, protagonist, …)   |
+| `ContextBuilder.forPack(id)`     | the service populates its own context and needs only the pack         |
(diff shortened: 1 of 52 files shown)


a-frazier (approved):

Pento95:
The new stylePrompt.ts — ContextBuilder.resolveTemplate() is public and already does the pack → default-pack → built-in chain. Any reason not to lean on it? The two versions already differ slightly (resolveTemplate takes the template if it exists, resolveStylePrompt needs non-empty content), so an empty style template in a custom pack behaves differently depending on which one reads it.

Also in there, the try wraps both lookups. If the custom pack lookup throws for a real reason (DB error, not "template missing"), we skip the default-pack fallback and land on the built-in — losing the global customization. Worth tightening the catch around just the first lookup?

aiService.generateSuggestions still takes promptContext and only logs it, never passes it on. You were already touching that signature — want to drop it while you're there?

And characterVault.svelte.ts:341 / scenarioVault.svelte.ts:285 pass an explicit undefined, which is right for the Vault since it's global. Maybe a one-liner comment so nobody "fixes" it later?

Pento95 (approved):

PR #450, opened 2026-08-16
Entry toolbar overflow menu, image model select, and orphaned checkpoints

Pento95:
On a phone the entry header laid out up to nine icon buttons in a single non-wrapping row, so the rightmost ones rendered outside the card. Below the `sm` breakpoint the secondary actions now fold into an overflow menu; narrate stays inline, copy moves into the menu, and the token badge is hidden there rather than removed.

The image model select no longer prepends a saved id that the provider does not list as if it were a real choice — it reports it under the select instead, and only once the fetch has actually returned something.

Separately: deleting an entry left its checkpoint behind. The branch panel kept offering to branch from it, and the resulting branch pointed at an entry the database no longer held ("Fork entry <id> not found in database"). Both deletion paths now prune those checkpoints, as they already did for chapters and embedded images.

## Summary by CodeRabbit

* **New Features**
  * Added a responsive “More actions” menu for story entries on smaller screens.
  * Added messaging when a previously selected image model is unavailable.
  * Clarified branch messages when checkpoint entries are still loading or have been deleted.

* **Bug Fixes**
  * Prevented branching from deleted story entries or removing entries that serve as branch points.
  * Automatically removed checkpoints associated with deleted entries, including during imports and retry restoration.
  * Preserved unavailable saved image-model selections without adding invalid options.
  * Improved error handling for failed deletions and retry restorations.


a-frazier (approved):

a-frazier:
### Code review

Found 2 issues:

1. `handleStopGeneration` restores lorebook activation data before the new entry-deletion guard runs, bypassing the ordering its own comment requires

`restoreFromBackup` places `assertEntriesRemovable` first, specifically because "a refusal that surfaced later would leave [the activation] restore applied with nothing undone":

https://github.com/AventurasTeam/Aventuras/blob/fc521a6c814015b1c6a602e3a8ba83af86d7abf3/src/lib/services/generation/RetryService.ts#L119-L126

But `handleStopGeneration` calls `restoreActivationData` directly before calling `restoreFromBackup`, so on this path the guard runs after the destructive overwrite instead of before it. If a branch has forked at/after `entryCountBeforeAction` when the user stops generation, activation data gets clobbered before the throw:

https://github.com/AventurasTeam/Aventuras/blob/fc521a6c814015b1c6a602e3a8ba83af86d7abf3/src/lib/services/generation/RetryService.ts#L247-L258

2. `BranchPanel`'s new orphaned-checkpoint filter can misreport a valid checkpoint as "entry deleted" during a branch switch

`getLatestCheckpoint()` filters out checkpoints whose `lastEntryId` isn't in `story.entries`. But `performBranchSwitch` updates `currentStory.currentBranchId` synchronously and only repopulates `story.entries` after an `await`, so during that window a checkpoint belonging to the newly-selected branch can appear orphaned. This is the same "not loaded yet" vs. "actually gone" ambiguity `forkPointTitle` in this same file was written to distinguish for a different button (see the comment there), but the new `createBranchTitle` doesn't have an equivalent "not loaded yet" case:

https://github.com/AventurasTeam/Aventuras/blob/fc521a6c814015b1c6a602e3a8ba83af86d7abf3/src/lib/components/branch/BranchPanel.svelte#L96-L108

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #452, opened 2026-08-16
Make the action choices side panel configurable

Selene29:
## Summary

- keep adventure action choices inline by default and make the responsive side panel opt-in
- group the story-width and action-choice-width controls under **Settings → Interface → Story Layout**
- persist the side-panel preference and preferred choice-column width, with responsive shrinking on smaller windows
- normalize malformed stored widths within a broad 16–160 rem safety envelope

### Responsive action choices side panel

Responsive action choices side panel

### Story layout settings

Story layout settings

## Verification

- `npm test` — 83 files, 1,193 tests passed
- `npm run check` — 0 errors, 0 warnings
- `npm run lint` — 0 errors (203 existing boundary warnings)
- focused side-panel settings tests — 2 passed

## Summary by CodeRabbit

* **New Features**
  * Added an optional side panel for adventure action choices on wide screens.
  * Added settings to enable the panel and adjust its width.
  * Improved responsive story and input layouts for wider displays.
* **Bug Fixes**
  * Panel width settings are now validated and constrained to supported limits.
* **Tests**
  * Added coverage for default settings and panel-width validation.

### DIFF
--- a/src/lib/components/settings/tabs/interface.svelte
+++ b/src/lib/components/settings/tabs/interface.svelte
@@ -281,7 +281,8 @@
           max={ACTION_CHOICES_PANEL_WIDTH.max}
           step={1}
           value={settings.uiSettings.actionChoicesPanelWidth}
-          onValueChange={(width) => settings.setActionChoicesPanelWidth(width)}
+          onValueChange={(width) => settings.previewActionChoicesPanelWidth(width)}
+          onValueCommit={(width) => settings.setActionChoicesPanelWidth(width)}
         />
         <div class="text-muted-foreground flex justify-between text-xs">
           <span>{ACTION_CHOICES_PANEL_WIDTH.min} rem</span>
--- a/src/lib/components/settings/tabs/interface.svelte
+++ b/src/lib/components/settings/tabs/interface.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import { onMount } from 'svelte'
   import { settings, STORY_WIDTH_OPTIONS } from '$lib/stores/settings.svelte'
+  import { ACTION_CHOICES_PANEL_WIDTH } from '$lib/stores/uiLayoutSettings'
   import { ui } from '$lib/stores/ui.svelte'
   import { database } from '$lib/services/database'
   import { grammarService } from '$lib/services/grammar'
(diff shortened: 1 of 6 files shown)


Pento95:
Thanks for this — the opt-in default and the clamping of persisted widths both look right. A few things I stumbled on in the layout; I may have misread the CSS, so please check before changing anything.

**The panel may never appear on common screen sizes.** The container query is `min-width: 88rem` (1408px), but the grid's own minimum is 49.5rem (32 + 16 + 1.5 gap). On a 1080p or 1440×900 laptop the toggle would turn on and nothing visible would change — the `<aside>` still renders, just unstyled, so it lands below the story exactly where the choices already were. Is 88rem deliberate?

**Enabling the panel looks like it widens the story text past the chosen Story Width.** Inside the query, `max-width` becomes `calc(var(--story-max-width) + var(--choice-panel-width) + 1.5rem)` and column 1 is `minmax(32rem, 1fr)`, so `1fr` absorbs whatever column 2 doesn't take. With Story Width at "Default" (48rem) and the panel at 32rem, if the choices only need ~20rem the text column ends up around 60rem. Would `max-width: var(--story-max-width)` on `.story-column` give you what you intended?

**Related: the text width seems to change on every turn.** `with-choice-column` follows `sidePanelLayoutEnabled` (ignores streaming) while the `<aside>` follows `showActionChoices` (doesn't). Since `minmax(16rem, var(--choice-panel-width))` isn't an `fr` track it sizes to content, so with no aside it should collapse to 16rem and the story column grows — then shrinks back when generation ends.

**One I couldn't check without running the app:** `container-type: inline-size` implies `contain: layout`, which makes `.story-view` a containing block for `position: fixed` descendants, not just absolute ones. Could you verify popovers and dropdowns inside `StoryEntry` still position correctly? If they portal to `body` it's a non-issue, but it wouldn't show up in `check`, `lint` or the tests either way.

Minor: the width slider writes to SQLite on every `onValueChange`, so one drag across 16–160 can be a hundred-plus `setSetting` calls — `onValueCommit` would avoid that.

Selene29:
@Pento95 Thanks — I checked these against the actual grid behavior. One item needed a fix; the rest are intentional layout behavior or do not reproduce from the current structure. The fix is now pushed to the PR branch in `0e06df45`.

- **Slider persistence:** dragging still updates the reactive width continuously, but SQLite is written only from `onValueCommit`. This also removes the possibility of an older asynchronous drag write completing after the final value.
- **88rem container query:** retained intentionally. Below that content width the choices remain inline; the setting itself says "On wide windows." A 1440px window with the sidebar open is therefore expected to stay inline, while a sufficiently wide content area gets two columns. I agree this threshold can be changed as a product decision, but it is not derived from the grid's absolute minimum because squeezing the story to 32rem and choices to 16rem would make the opt-in layout technically fit but substantially less useful.
- **Story Width:** the first column does not absorb the choices column's configured width in the available layout. The second explicit track has a fixed growth limit of `var(--choice-panel-width)`, and the wrapper max is Story Width + choice width + gap, leaving the story at its selected maximum when that space is available.
- **Turn-to-turn width:** the explicit second grid track does not collapse merely because the aside is temporarily absent. `with-choice-column` intentionally follows the enabled layout rather than streaming state, so the empty track remains during generation and prevents the story/input columns from jumping.
- **Container containment:** this is a valid general caution. In the current StoryView tree, the StoryEntry popover content is portalled and I found no fixed-position descendant that remains inside the query container, so I made no speculative code change here.

Verification on the pushed branch: 1,193 tests passed, `svelte-check` reported 0 errors/0 warnings, and the modified Svelte component analyzer was clean.

PR #458, opened 2026-08-20
Add branch-aware story landmark navigation

collegatore:
## Summary

- Add a persistent story navigation panel with entry-number and full-lineage landmark navigation.
- Label landmarks with their owning branch and support inline checkpoint renaming.
- Add a landmark navigation mode that stays on the current branch or switches to the owning branch before jumping.

## Verification

- npm run check — passed with 0 errors and 0 warnings.
- npm run lint — passed with 0 errors (existing warnings only).
- npm test — 1120 tests passed.

Target branch: upstream/master.

## Summary by CodeRabbit

* **New Features**
  * Added story navigation with entry numbers, branch landmarks, checkpoints, renaming, and responsive controls.
  * Added mobile swipe gestures, desktop resizing, persistent panel preferences, and header/mobile-menu access.
  * Added automatic branch-tree expansion and improved response actions across screen sizes.

* **Bug Fixes**
  * Improved story loading, navigation, branch switching, and cleanup reliability.
  * Added clear errors for invalid navigation, failed renames, and loading failures.
  * Prevented conflicting edits or deletions during generation and retries.
  * Improved handling of unavailable checkpoint data.

### DIFF
--- a/src/lib/components/layout/AppShell.svelte
+++ b/src/lib/components/layout/AppShell.svelte
@@ -4,6 +4,7 @@
   import { story } from '$lib/stores/story.svelte'
   import { settings } from '$lib/stores/settings.svelte'
   import Sidebar from './Sidebar.svelte'
+  import StoryNavPanel from './StoryNavPanel.svelte'
   import Header from './Header.svelte'
   import ProfileWarningBanner from './ProfileWarningBanner.svelte'
   import StoryView from '$lib/components/story/StoryView.svelte'
@@ -225,27 +226,54 @@
       ></div>
     {/if}
 
+    <!-- Left edge swipe zone for opening the nav panel — a region-owning gesture, the mirror
+         of the right-edge zone above. Mounted only while neither panel is open, so a
+         right-swipe that belongs to the sidebar's tab strip or to dismissing a panel cannot
+         also open this one. -->
+    {#if !ui.sidebarOpen && !ui.navPanelOpen && story.currentStory}
+      <div
+        class="swipe-edge-zone swipe-edge-zone-left"
+        use:swipe={{ onSwipeRight: () => ui.toggleNavPanel(), threshold: 30 }}
+      ></div>
+    {/if}
+
     <!-- Main content area -->
     <div class="flex flex-1 flex-col overflow-hidden">
       <Header />
 
-      <main class="flex-1 overflow-hidden">
-        {#if ui.activePanel === 'story' && story.currentStory}
-          <StoryView />
-        {:else if ui.activePanel === 'gallery' && story.currentStory}
-          <GalleryTab />
-        {:else if ui.activePanel === 'lorebook' && story.currentStory}
-          <LorebookView />
-        {:else if ui.activePanel === 'memory' && story.currentStory}
-          <MemoryView />
-        {:else if ui.activePanel === 'vault'}
-          <VaultPanel />
-        {:else if ui.activePanel === 'library' || !story.currentStory}
-          <LibraryView />
-        {:else if children}
-          {@render children()}
+      <!-- `relative` so the navigation panel can overlay the content area without the header:
+           it must take no layout space, or opening it would narrow the story column and
+           reflow the text the reader is looking at. -->
+      <div class="relative min-h-0 flex-1">
+        {#if ui.navPanelOpen && story.currentStory}
+          <button
+            class="nav-panel-scrim"
+            onclick={() => ui.closeNavPanel()}
+            aria-label="Close story navigation"
+          ></button>
+          <div class="nav-panel-container">
+            <StoryNavPanel />
+          </div>
         {/if}
-      </main>
+
+        <main class="h-full overflow-hidden">
+          {#if ui.activePanel === 'story' && story.currentStory}
+            <StoryView />
+          {:else if ui.activePanel === 'gallery' && story.currentStory}
+            <GalleryTab />
+          {:else if ui.activePanel === 'lorebook' && story.currentStory}
+            <LorebookView />
+          {:else if ui.activePanel === 'memory' && story.currentStory}
+            <MemoryView />
+          {:else if ui.activePanel === 'vault'}
+            <VaultPanel />
+          {:else if ui.activePanel === 'library' || !story.currentStory}
+            <LibraryView />
+          {:else if children}
+            {@render children()}
+          {/if}
+        </main>
+      </div>
     </div>
 
     <!-- Sidebar (Right aligned) -->
@@ -395,6 +423,32 @@
     outline-offset: -2px;
   }
 
+  /*
+   * Story navigation panel — an overlay at every width, unlike the right sidebar. Taking
+   * layout space would narrow the story column and reflow its text, moving the reader's
+   * place before they have navigated anywhere.
+   */
+  .nav-panel-container {
+    position: absolute;
+    left: 0;
+    top: 0;
+    bottom: 0;
+    width: 16rem;
+    z-index: 45;
+  }
+
+  /* Desktop keeps the story visible and interactive behind the panel: the reader jumps,
+     looks at where they landed, and jumps again without reopening anything. */
+  .nav-panel-scrim {
+    display: none;
+    position: absolute;
+    inset: 0;
+    background-color: rgba(0, 0, 0, 0.5);
+    z-index: 44;
+    border: none;
+    cursor: pointer;
+  }
+
   /* Right edge swipe zone */
   .swipe-edge-zone {
     position: fixed;
@@ -403,6 +457,11 @@
     width: 20px;
     height: 100%;
     z-index: 30;
+  }
+
+  .swipe-edge-zone-left {
+    right: auto;
+    left: 0;
   }
 
   /* Mobile styles */
@@ -424,11 +483,30 @@
       width: 30px;
       top: max(var(--safe-top), 28px);
     }
+
+    .nav-panel-scrim {
+      display: block;
+    }
+
+    .nav-panel-container {
+      width: calc(100vw - 3rem);
+      max-width: 288px;
+      animation: slide-in-left 0.2s ease-out;
+    }
   }
 
   @keyframes slide-in {
     from {
       transform: translateX(100%);
+    }
+    to {
+      transform: translateX(0);
+    }
+  }
+
+  @keyframes slide-in-left {
+    from {
+      transform: translateX(-100%);
     }
     to {
       transform: translateX(0);
--- a/src/lib/components/layout/AppShell.svelte
+++ b/src/lib/components/layout/AppShell.svelte
@@ -276,7 +276,7 @@
                  nothing to resize. -->
             <button
               type="button"
-              class="resizer-handle nav-panel-resizer hidden h-full sm:block"
+              class="resizer-handle nav-panel-resizer h-full"
               onmousedown={(e) => startResizing('navPanel', e)}
               aria-label="Story Navigation Resizer"
             ></button>
@@ -511,6 +511,12 @@
 
     .nav-panel-scrim {
       display: block;
+    }
+
+    /* The panel overlays at a fixed width here, so there is nothing to drag. Tied to this
+       query rather than a utility class, so it cannot disagree with the overlay breakpoint. */
+    .nav-panel-resizer {
+      display: none;
     }
 
     .nav-panel-container {
--- a/src/lib/components/layout/AppShell.svelte
+++ b/src/lib/components/layout/AppShell.svelte
@@ -123,26 +123,37 @@
     clampDebugButtonToViewport()
   }
 
-  // Resizing logic
-  let isResizing = $state(false)
+  // Resizing logic — two panels, one drag at a time
+  let resizing = $state<'sidebar' | 'navPanel' | null>(null)
+  const isResizing = $derived(resizing !== null)
+  let navPanelEl = $state<HTMLDivElement | null>(null)
 
-  function startResizing(e: MouseEvent) {
-    isResizing = true
+  function startResizing(target: 'sidebar' | 'navPanel', e: MouseEvent) {
+    resizing = target
     e.preventDefault()
   }
 
   function stopResizing() {
-    if (!isResizing) return
-    isResizing = false
+    if (!resizing) return
+    const target = resizing
+    resizing = null
     // Persist the final width to the database now that resizing is complete.
-    settings.setSidebarWidth(settings.uiSettings.sidebarWidth)
+    if (target === 'sidebar') {
+      settings.setSidebarWidth(settings.uiSettings.sidebarWidth)
+    } else {
+      settings.setNavPanelWidth(settings.uiSettings.navPanelWidth)
+    }
   }
 
   function handleMouseMove(e: MouseEvent) {
-    if (!isResizing) return
+    if (!resizing) return
 
-    // Sidebar is on the right, so width is window.innerWidth - mouseX
-    const newWidth = window.innerWidth - e.clientX
+    // The sidebar is flush to the right edge; the nav panel starts wherever the content
+    // area does, which is not the window edge once a safe-area inset is in play.
+    const newWidth =
+      resizing === 'sidebar'
+        ? window.innerWidth - e.clientX
+        : e.clientX - (navPanelEl?.getBoundingClientRect().left ?? 0)
 
     // Constraints
     if (
@@ -150,7 +161,11 @@
       newWidth <= Math.min(MAX_SIDEBAR_WIDTH, window.innerWidth * MAX_SIDEBAR_RATIO)
     ) {
       // For performance, update the reactive state directly without saving to the database on every mouse movement.
-      settings.uiSettings.sidebarWidth = newWidth
+      if (resizing === 'sidebar') {
+        settings.uiSettings.sidebarWidth = newWidth
+      } else {
+        settings.uiSettings.navPanelWidth = newWidth
+      }
     }
   }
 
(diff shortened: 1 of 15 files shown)


collegatore:
Addressed in 1aff6a07:

- Guarded the Branch panel's ancestor walk against cycles.
- Added feedback for invalid entry navigation and failed checkpoint renames.
- Catch story-load failures, return to the library, and show the error.
- Ignore invalid persisted navigation-panel widths.
- Apply mobile defaults before publishing the loaded story.
- Refactored landmark navigation into a native button, with rename controls as separate siblings.

Intentionally unchanged:

- Landmark navigation mode resets to the safe current-branch default when the panel remounts.
- Navigation-panel resizing continues to use the shared sidebar constraints; no observed layout fault warrants separate limits yet.
- CodeRabbit's docstring warning and ESLint path failure are external review-environment configuration issues; local checks pass.

Validation: npm run check (0 errors/warnings), npm test (1,210 passed), npm run lint (0 errors; 203 existing warnings).

collegatore:
Addressed the follow-up CodeRabbit feedback in a69d5250.\n\n- Made story loads serialized and last-request-wins, with stale-load guards.\n- Clamped persisted and saved navigation-panel widths to the shared 250–800 range.\n- Reset All now persists the default navigation-panel width and closed state.\n- Return to the library if opening an imported story fails.\n- Shortened the openStory comment.\n\nValidated with 
pm run check, 
pm test (1,210 passing), and 
pm run lint.

src/lib/components/layout/AppShell.svelte
```diff
@@ -47,6 +48,14 @@
     }
   }
 
+  // A swipe that closes the nav panel bubbles up to here, where it would otherwise be read
+  // as the gesture that opens the sidebar. Opening the nav panel needs no such guard: its
+  // edge zone is only mounted when nothing else claims the gesture.
+  function handleSwipeLeftOnShell() {
```
a-frazier: **Swipe-left guard fires too late — dismissing the nav panel also opens the sidebar.**

Nav panel open on a touch device, user swipes left on the panel to dismiss it:

1. `StoryNavPanel`'s `use:swipe` (`StoryNavPanel.svelte:115`) calls `ui.closeNavPanel()`, setting `ui.navPanelOpen = false` synchronously.
2. The same `touchend` bubbles to `.app-shell`'s swipe handler.
3. `handleSwipeLeftOnShell` runs `if (ui.navPanelOpen) return` — now `false` — so it falls through to `handleSwipeLeft()` → `ui.toggleSidebar()`.
4. The right sidebar opens and persists `sidebar_open=true`.

The comment says this guard exists precisely to prevent that, but because the inner handler already flipped the flag, it never fires.

**Suggested direction:** decide the gesture's target from state captured at gesture *start* (e.g. on `touchstart`), or have the shell handler bail when the event originated inside the nav-panel subtree, rather than relying on a store flag the inner handler mutates first.

src/lib/components/layout/StoryNavPanel.svelte
```diff
@@ -0,0 +1,270 @@
...
+    goTo(entry.id, `Jumped to entry ${entryNumber(entry)}`)
+  }
+
+  function setLandmarkNavigationMode(value: string) {
+    if (value === 'current-branch' || value === 'checkpoint-branch') {
+      landmarkNavigationMode = value
+    }
+  }
+
+  async function goToLandmark(landmark: Landmark) {
```
a-frazier, line 79: **`goToLandmark` in `checkpoint-branch` mode can land on the branch's fork point while still toasting "Jumped to entry N".**

The nav panel can be open while `ui.activePanel !== 'story'` (lorebook / gallery / memory on desktop). User clicks a landmark on another branch with mode set to *Switch to checkpoint branch*:

1. `goToLandmark` awaits `story.switchBranch(landmark.branchId)`, which emits `BranchSwitched` while `activePanel !== 'story'` → StoryView's subscription sets `pendingBranchLanding = true`.
2. `goTo` calls `ui.requestEntryScroll(landmark.entryId)` then `ui.setActivePanel('story')`.
3. StoryView mounts; its panel effect (`StoryView.svelte:415`) calls `ui.consumeEntryScroll()` unconditionally (clearing the request), sees `pending` is true, runs `performBranchLanding()` and returns.
4. The entry-scroll request is discarded. The reader lands on the fork point, not the clicked landmark — but the toast reports success.

The precedence of a pending branch landing over an entry-scroll request is pre-existing and intentional (`StoryView.svelte:419-434`, comment at :428); it is correct for its original caller (the Branches panel, which switches without an entry-scroll request). `goToLandmark` is the first caller to queue both from a non-story panel.

**Suggested direction:** after a branch switch, land the landmark entry explicitly rather than delegating to StoryView's mount effect, or suppress the `pendingBranchLanding` path when an explicit entry-scroll request is also queued. At minimum, don't show the success toast when the jump degraded to a fork-point landing.

src/lib/stores/story.svelte.ts
```diff
@@ -544,21 +550,42 @@ class StoryStore {
...
     clearImageMarkerCache()
 
+    // A remembered desktop panel must not briefly cover a story as it starts loading on a
+    // narrow display. Do this before publishing `currentStory`, which mounts those panels.
+    ui.setMobileDefaults()
     this.currentStory = story
-    this.currentBgImage = await database.getBackgroundForBranch(storyId, story.currentBranchId)
+    const backgroundImage = await database.getBackgroundForBranch(storyId, story.currentBranchId)
+    if (seq !== this.storyLoadSeq) return
+    this.currentBgImage = backgroundImage
```
a-frazier, line 621: **A superseded `performStoryLoad` can leave the store half-updated.** (Low severity — see caveat below.)

This call sets `this.currentStory = story` (line 585), then during `await database.getBackgroundForBranch(...)` a newer `loadStory` / `closeStory` / `clearCurrentStory` bumps `storyLoadSeq`, so it bails at line 587 — before `reloadEntriesForCurrentBranch`. Normally the newer load repairs the store. But if that newer load itself throws early (e.g. `database.getStory` returns `null` → `Story not found`, which a sync-removed or just-deleted story hits), its caller catches and shows the library while `story.currentStory` is left as the half-loaded story with stale `entries` / `branches` / `characters`.

**Caveat:** for an in-process single-user SQLite DB this is a narrow edge. The bad state is in-memory only, self-heals on the next successful load, and needs the supersession to coincide with an early failure of the superseding load. "Won't fix" is defensible. Flagging it because the fix is cheap and local to a function this PR is already rewriting.

**Suggested direction:** publish `currentStory` together with the dependent collections in one assignment after the last `await`, or roll `currentStory` back on the abort branch.

a-frazier (commented):
Code review (high effort, with a pass for over-engineering). Three findings, all inline. All three are introduced by this branch rather than pre-existing on `master`. The over-engineering pass turned up nothing actionable. Nothing here is a hard crash.

a-frazier (changes requested):
The items posted before the recent rebase are all still valid.

src/lib/components/layout/AppShell.svelte
collegatore: Swipe gestures are broken at several levels because of appshell registration; I just made it minimally viable here. Here's a focused PR fix: #472

src/lib/components/layout/StoryNavPanel.svelte
collegatore: The symptom is real but the traced mechanism doesn't reproduce, so the fix went elsewhere.

Step 1 requires StoryView's `BranchSwitched` subscriber to run while `activePanel !== 'story'`. `<StoryView />` is inside `{#if ui.activePanel === 'story' && story.currentStory}` in `AppShell`, so it is destroyed rather than hidden — the subscription is torn down and the event reaches nobody. `pendingBranchLanding` is also a per-instance `let`, so the remount starts it `false` regardless. Walking the scenario with that correction, the mount effect takes the `entryId` path and lands on the clicked landmark.

The race is in the configuration the report doesn't cover: the panel used **with the story on screen**, which the desktop push layout makes the common case. `switchBranch` emits `BranchSwitched` from inside its own await, so the view starts `performBranchLanding()` (anchor to the end of the branch) and then `goTo` queues the landmark jump — two render windows and two scrolls, with the winner resting on microtask/RAF ordering.

`goToLandmark` now claims the landing before the switch, and the subscriber stands down when it sees the claim. The claim is released in a `finally`: only a mounted story view consumes it, so on the other-panel path nothing would take it and it would otherwise suppress an unrelated switch later.

The toast point stands on its own and is fixed: it now checks the same condition the view checks before landing, and reports that the entry is not in this branch rather than asserting a jump.

Fixed in 25cf7faa. Not yet exercised in the running app — it is lifecycle timing, and the suite has no DOM.

src/lib/stores/story.svelte.ts
collegatore: Confirmed, and it bites this PR harder than the caveat suggests: the navigation panel is gated on `currentStory`, so in that window it mounts and runs `buildLandmarks(story.entries, …, activeBranch)` — one story's entries against another story's branch. Every entry number in the list would be wrong, and a jump would target the wrong story's entry.

Went with the rollback rather than the single-assignment variant, because `reloadEntriesForCurrentBranch()` reads `this.currentStory` — which is why it is published early. All seven post-publish bail-outs now unpublish on the way out. Loads are serialised, so the load that superseded this one has not started and this cannot clear a story it already published; if that load then fails early, the reader gets the library instead of the mismatch.

Fixed in 25cf7faa.

src/lib/components/story/LibraryView.svelte
```diff
@@ -33,10 +33,20 @@
...
   async function openStory(storyId: string) {
     ui.resetScrollBreak()
-    await story.loadStory(storyId)
     ui.setActivePanel('story')
+    try {
+      await story.loadStory(storyId)
+    } catch (error) {
+      ui.setActivePanel('library')
+      ui.showToast(errMessage(error), 'error')
+    }
```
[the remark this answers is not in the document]
collegatore: Request-scoping landed in a69d5250 (`storyLoadSeq` plus a serialising `storyLoadChain`), a day after this comment: a superseded load returns before publishing anything.

The second half — a stale *rejection* reaching `openStory`'s catch and returning the reader to the library — survived that, since not every await sits behind a seq check. d019fd46 swallows rejections from a load the reader has already replaced; current ones still propagate.

src/lib/components/layout/AppShell.svelte
```diff
@@ -225,27 +241,66 @@
...
+            class="nav-panel-container relative flex"
+            style:width="{settings.uiSettings.navPanelWidth}px"
+          >
+            <StoryNavPanel />
+            <!-- Desktop only: on mobile the panel overlays at a fixed width, so there is
+                 nothing to resize. -->
+            <button
+              type="button"
+              class="resizer-handle nav-panel-resizer hidden h-full sm:block"
+              onmousedown={(e) => startResizing('navPanel', e)}
```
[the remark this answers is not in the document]
collegatore: Fixed in 333307d1 — the resizer's visibility now lives in the same `@media (max-width: 768px)` block as the overlay rather than in a `sm:` utility (640px), so the two cannot disagree about where dragging is possible.

src/lib/components/layout/StoryNavPanel.svelte
```diff
@@ -0,0 +1,284 @@
...
+                </span>
+                <input
+                  type="text"
+                  class="input min-w-0 flex-1 px-1 py-0.5 text-sm"
+                  bind:value={renameValue}
+                  onkeydown={(e) => {
+                    if (e.key === 'Enter') confirmRename()
+                    if (e.key === 'Escape') cancelRename()
+                  }}
+                />
```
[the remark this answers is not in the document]
collegatore: Fixed in 333307d1 — added `aria-label="Checkpoint name"` to the rename input.

src/lib/components/layout/StoryNavPanel.svelte
```diff
@@ -0,0 +1,284 @@
...
+                  <!-- Wrapped rather than truncated: a name the reader chose is the only thing
+                       telling these rows apart, and a touch device has no tooltip to fall back
+                       on. The list is a handful of rows, so the vertical space is affordable. -->
+                  <span class="text-surface-200 block text-sm break-words">{landmark.label}</span>
+                  <span class="text-surface-500 block truncate text-xs">{landmark.branchName}</span>
+                </span>
+              </button>
+              {#if landmark.checkpointId}
+                <button
+                  class="text-surface-500 hover:text-surface-200 absolute top-1 right-1 flex min-h-[32px] min-w-[32px] items-center justify-center p-1 transition-opacity sm:min-h-0 sm:min-w-0 sm:p-0.5 sm:opacity-0 sm:group-hover:opacity-100"
```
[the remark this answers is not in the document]
collegatore: Fixed in 333307d1 — the rename button now also reveals itself on `sm:focus-visible`, so it is reachable by keyboard and not only by hovering pointer.

src/lib/components/layout/AppShell.svelte
collegatore: Rebased on #472 after it was merged, so work is synced

src/lib/stores/story.svelte.ts
```diff
@@ -544,21 +550,71 @@ class StoryStore {
...
+
+  private async performStoryLoad(storyId: string, seq: number): Promise<void> {
+    try {
+      await this.runStoryLoad(storyId, seq)
+    } catch (error) {
+      // A load the reader has already replaced must not report its failure. Its caller sends
+      // them back to the library on error, which would discard the story they switched to and
+      // explain it with the name of the one they left. The newer load speaks for itself.
+      if (seq !== this.storyLoadSeq) return
+      throw error
```
a-frazier, line 591: **Non-superseded load failure leaves the story store half-open.**

When a load fails *after* `this.currentStory = story` is published (~line 617) but has not been superseded, `runStoryLoad` throws, `seq === this.storyLoadSeq` is still true, and this `throw error` propagates. `openStory` / `triggerImport` catch it and call `ui.setActivePanel('library')` + a toast — but nothing resets the story store. `currentStory` now names the new story while `entries` / `branches` / world state are still the previous story's or partial.

The `supersededAfterPublish(seq)` guards handle the superseded case by nulling `currentStory`; this plain throw path does not. AppShell renders `LibraryView` while `activePanel === 'library'`, so the mismatch isn't visible immediately, but it stays latent in the store until the next `loadStory` / `closeStory`, and any background consumer of `currentStory` + `entries` in that window sees a cross-story mismatch.

Likelihood is low with in-proc SQLite — a mid-load read would need `SQLITE_BUSY` under a concurrent writer, a Windows AV file-lock hiccup, or a permanently-malformed row — but the fix is small; mirror the superseded path:

```ts
} catch (error) {
  if (seq !== this.storyLoadSeq) return
  if (this.currentStory?.id === storyId) this.closeStory()
  throw error
}
```

src/lib/components/layout/StoryNavPanel.svelte
```diff
@@ -0,0 +1,285 @@
...
+      landmarkNavigationMode = value
+    }
+  }
+
+  async function goToLandmark(landmark: Landmark) {
+    const currentBranchId = story.currentStory?.currentBranchId ?? null
+    if (landmarkNavigationMode === 'checkpoint-branch' && currentBranchId !== landmark.branchId) {
+      // Claimed before the switch, because the event that triggers the story view's own
+      // end-of-branch landing is emitted inside it.
+      ui.claimBranchLanding()
```
a-frazier: **`branchLandingClaimed` is an un-keyed boolean that an unrelated in-flight `switchBranch` can consume.**

`claimBranchLanding()` sets a single process-wide boolean, then `await story.switchBranch(...)`. If a *different* branch switch is already in flight when the claim is set, that earlier switch reaches `eventBus.emit('BranchSwitched')` (`story.svelte.ts:4050`) — which has no seq re-check after its top-of-function guard — while the claim is set. StoryView's subscriber calls `consumeBranchLandingClaim()` and the claim is spent on the wrong switch: the in-flight switch's own end-of-branch landing is suppressed, and this landmark switch's landing then runs unclaimed, racing `goTo()`'s `requestEntryScroll`.

Repro: Branches sidebar + nav panel both open on desktop; click a branch in BranchPanel, then click a checkpoint-branch landmark before the first switch settles. Also reachable with two quick landmark clicks, or a landmark click right after `createBranchFromCheckpoint` (which calls `switchBranch` internally).

Impact is cosmetic — a one-frame "land at end of branch, then jump to the entry" flash that self-corrects; the queue's last-one-wins guard keeps the final branch state correct. Keying the claim to the target `branchId`, or having `performBranchSwitch` re-check `seq` immediately before it emits, closes the window.

src/lib/stores/story.svelte.ts
```diff
@@ -1054,10 +1113,17 @@ class StoryStore {
 
     // Legacy behavior: delete just this one entry (no world state changes)
     const droppedCheckpoints = this.checkpointsAnchoredTo(new Set([entryId]))
+    // Chapters hold foreign keys to the entries at their boundaries, with no ON DELETE
+    // behaviour, so a boundary entry cannot go without them — same as the cascade path.
+    const chaptersToDelete = this.chapters.filter(
```
a-frazier: **Single-entry delete of a chapter boundary silently destroys the chapter and its summary.**

This makes the legacy (non-rollback) `deleteEntry` path delete any chapter whose `startEntryId` / `endEntryId` is the deleted entry. That row carries the AI-generated `summary` (a model call, and a context-injection input) — discarded with no toast and no confirmation naming the chapter.

The comment says "same as the cascade path," but the two aren't equivalent. `deleteEntriesFromPosition` only ever deletes a contiguous *suffix* of entries, so it only drops a trailing run of chapters — the entries they covered go with them, and `chapters` stays `[1..k]`. Applying the same filter to a single-entry delete can remove a *middle* chapter: delete chapter 2's boundary out of `[1,2,3,4]` and you're left with `chapters = [1,3,4]`, entries formerly in chapter 2 now un-chapterized between chapter 1 and chapter 3. `validateChapterIntegrity` does not repair that (the survivors still have valid refs), so number-based labelling / sorting stays wrong until re-chapterization. (`getNextChapterNumber` uses `max+1`, so at least no number collision.)

Pre-PR this delete hit the `chapters.start_entry_id` / `end_entry_id` FK (migration `002_chapters_checkpoints.sql`, enforced via `PRAGMA foreign_keys = ON` in `database.ts`) and threw — a loud, safe no-op. This trades that for a silent lossy success.

Suggested fix: when the deleted entry is a chapter boundary, re-point the boundary to the adjacent surviving entry and decrement `entryCount`, deleting the chapter only if it would become empty — so a single-entry delete shrinks a chapter instead of destroying it. At minimum, a confirmation that names the affected chapter(s).

a-frazier (commented):
Three findings from reviewing the rebased branch. All three are in the delete / story-load concurrency work; none block on their own but F3 is genuine silent data loss worth addressing before merge.

src/lib/components/layout/StoryNavPanel.svelte
a-frazier: Whoops.  I didn't mean to post this one.  It's really minor.

src/lib/stores/story.svelte.ts
collegatore: Fixed in 4c193a0c — a failure after `currentStory` is published now tears the store down (`closeStory()`) before the error reaches the caller, matching what the superseded path does.

src/lib/components/layout/StoryNavPanel.svelte
collegatore: Fixed in 4c193a0c — the claim now carries its target `branchId` and is only consumed when the emitted `BranchSwitched` matches it, so a switch already in flight can no longer spend it; the caller clears it unconditionally in `finally`.

src/lib/stores/story.svelte.ts
collegatore: You're right, and the "same as the cascade path" reasoning was wrong — that path only ever drops a trailing run. 4c193a0c refuses the delete instead, naming the chapter, so the summary can't be lost silently. I did not implement the boundary re-point: `updateChapter` can't write `startEntryId`/`endEntryId`/`entryCount`, so doing it safely means extending the delete transaction, which I'd rather land as its own reviewed change than add unverified here.

a-frazier (approved):

PR #462, opened 2026-08-21
feat: add setting to change prompt-pack for a story

proAlexandr:
## Problem

Stories imported from `.avt` files cannot specify which prompt pack they should use. This makes it impossible to use custom prompt packs with imported stories.

## Change

Adds a prompt pack selector to Story Settings, allowing a different prompt pack to be selected for an imported story.

## Verification

- pre-commit hooks: `npm run lint` and `npm run check`
- Import an `.avt` story, select a custom prompt pack in Story Settings, and verify that the next generation uses it.

Screenshot 2026-08-21 at 17 11 53
Screenshot 2026-08-21 at 17 11 59
Screenshot 2026-08-21 at 17 12 09

## Summary by CodeRabbit

* **New Features**
  * Added a prompt-pack selector to story settings.
  * Automatically loads available prompt packs and the story’s current selection.
  * Saves prompt-pack changes with loading and error feedback.
  * Uses the selected prompt pack when resolving story prompts.

### DIFF
--- a/src/lib/components/settings/tabs/story-settings.svelte
+++ b/src/lib/components/settings/tabs/story-settings.svelte
@@ -10,7 +10,10 @@
   } from '$lib/services/prompts/templates'
   import { ContextBuilder } from '$lib/services/context'
   import { database } from '$lib/services/database'
+  import { packService } from '$lib/services/packs/pack-service'
+  import type { PresetPack } from '$lib/services/packs/types'
   import WritingStyleFields from '$lib/components/shared/WritingStyleFields.svelte'
+  import * as Select from '$lib/components/ui/select'
   import { Textarea } from '$lib/components/ui/textarea'
   import { Button } from '$lib/components/ui/button'
   import { Label } from '$lib/components/ui/label'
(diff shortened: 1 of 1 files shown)


collegatore:
I actually addressed the issue in PR #446. Do you think there is anything amiss there?

a-frazier:
@collegatore Your PR only displays the prompt pack. One of this PR's features is giving the user a UX to change it. 

The feature needs rebuilt because your PR added a lot of infrastructure that can be leveraged.

collegatore:
@a-frazier I meant that your stated goal - to allow an imported story to resolve into a custom pack - is there. It acts on the import level, not in the settings afterwards. There's a toggle in settings to add a new behavior to old exports - those that do not carry any pack info at all in their metadata will actively ask to be assigned. The new export-import cycles get this automatically. Your approach is more general but also more complicated, since it implies that any pack may be replaced with any other, regardless of consequences on runtime or pack variables. This is the question that stopped me from going this path initially, but I'll be glad to see it implemented.

proAlexandr:
> I actually addressed the issue in PR #446. Do you think there is anything amiss there?

Hi,
I agree that #446 solves my issue.
@a-frazier can you please advice (as a repo collaborator) on the next step. Do you prefer me to update the PR so we can eventually merge it? Or do you prefer to close it? Both options works for me. Thanks!

a-frazier:
> @a-frazier can you please advice (as a repo collaborator) on the next step. Do you prefer me to update the PR so we can eventually merge it? Or do you prefer to close it? Both options works for me. 

@proAlexandr If you'd like to work on the Story Setting to change the active prompt pack, I'd recommend opening a new PR -- I'm going to close this one for now.

If you'd rather not because the reason you wanted the feature is already taken care of another way, I totally understand that too.  😂

PR #468, opened 2026-08-26
Android build hygiene: minSdk 26, lint in CI, resource cleanup

Pento95:
Android build hygiene, split out of the review on #467 so that PR stays focused on the feature.

## 1. `minSdk` 24 → 26 — fixes a real crash

`MainActivity.onWebViewCreate` calls `WebView#setRendererPriorityPolicy`, added in API 26, with no version guard. Unlike an inlined constant, a missing *method* resolves at runtime: on Android 7.0/7.1 that line throws `NoSuchMethodError` as soon as the WebView opens, i.e. at startup. `minSdk = 24` was promising a compatibility the code never had.

Android lint confirms it as an **error**, not a warning:

```
MainActivity.kt:44: Error: Call requires API level 26 (current min is 24):
  android.webkit.WebView#setRendererPriorityPolicy [NewApi]
```

Raising `minSdk` also removes the `SDK_INT >= O` guard around `createNotificationChannel` (now dead code, along with the `android.os.Build` import) and clears both `InlinedApi` warnings — `RENDERER_PRIORITY_IMPORTANT` and, from #467, `IME_FLAG_NO_PERSONALIZED_LEARNING`.

The cost is dropping Android 7.0/7.1: under 1-2% of installed devices and, per the above, never actually working.

## 2. Android lint in CI

CI built the APK but never linted it, which is why the error above went unnoticed. The step runs after the build, since lint needs the Rust libraries and the wry-generated sources. On failure the HTML report is uploaded as an artifact.

Two checks inherited from the `tauri android init` template are disabled — `MissingTvBanner` and `ImpliedTouchscreenHardware`. This app is not a Leanback app and cannot be used without a touchscreen.

## 3. Unused template resources and the hdpi icon

Also from the template, never referenced: `layout/activity_main.xml`, the seven placeholder colours, `drawable/ic_launcher_background.xml`, `drawable-v24/ic_launcher_foreground.xml`, and `ic_launcher_round` across five densities — the manifest declares `android:icon` only, never `android:roundIcon`. Kept: `values/ic_launcher_background.xml` (the colour the adaptive icon uses) and the `ic_launcher_foreground` PNGs. Emptying `drawable-v24` takes the folder and its `ObsoleteSdkInt` warning with it.

`mipmap-hdpi/ic_launcher.png` was 49x49 px where hdpi wants 72x72, so it rendered at 33dp while every other density gave 48dp — visibly small on hdpi screens. Resampled from the xxxhdpi original.

Raising `minSdk` made the `-v26` on `mipmap-anydpi-v26` redundant in turn, so the folder is now plain `mipmap-anydpi`. AAPT2 stamps it back to `-v21`, the minimum `anydpi` itself requires, and `mipmap/ic_launcher` still resolves to the adaptive XML.

Left alone on purpose: `MonochromeLauncherIcon`. `ic_launcher_foreground` is 100% opaque, so reusing it as the monochrome layer would render a filled square — worse than the warning. That needs a silhouette asset drawn for the purpose.

## 4. `compileApk.sh`: one more JDK location

The script scanned `/usr/lib/jvm` and the Android Studio JBR. Distros that no longer package an LTS JDK (Fedora 41+ ships only the latest release) force the JDK into `$HOME`, so `$HOME/.local/lib/jvm/*/` is now scanned too. The existing `jdk_ok()` filter is unchanged.

## Verified

Locally: debug APK builds, the APK carries every icon density and no `ic_launcher_round`, and `./gradlew :app:lintUniversalDebug` goes from **3 errors, 28 warnings** to **0 errors, 12 warnings**. What remains is outdated androidx dependencies, two findings in wry-generated sources, and the monochrome tag above — none of them fail the build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android build validation by checking keyboard injection after APK creation.
  * Added lint verification and failure-report uploads to make build issues easier to identify.

* **Build Improvements**
  * Expanded automatic JDK detection to support installations stored in the user’s local JVM directory.
  * Improved build reliability across additional development environment configurations.


a-frazier (approved):

a-frazier (approved):
We definitely don't need the deleted launcher icons?

PR #473, opened 2026-09-01
Feat/generation context data source

failerko:
## Summary by CodeRabbit

* **Improvements**
  * Generation phases now use more complete story context, including recent turns and scene details.
  * Prompt buffers retrieve the appropriate story entries beyond the currently loaded window.
  * Templates receive only the context they use, improving consistency and efficiency.
  * Added safeguards for mismatched story data and duplicate entry positions.

* **Performance**
  * Faster story-entry lookup supports more responsive prompt generation.

* **Bug Fixes**
  * Improved turn classification and suggestion generation across branches and story states.


lib/pipeline/definitions/generation-context.ts
```diff
@@ -128,30 +130,130 @@ const floorThreads = (rows: readonly ThreadRow[] | undefined): FloorThread[] =>
...
+ * and `position` is the handle a template wants.
+ */
+function promptEntry(entry: StoryEntry) {
+  return { position: entry.position, content: promptProse(entry) }
+}
+
+// Scene state is written by classification, so it lives on AI-authored rows; a
+// user_action only inherits it forward. Reading the AI row directly keeps the
+// prompt's scene independent of which kind happens to sit at the tail.
+async function readSceneSource(db: DbCtx['db'], branchId: string): Promise<StoryEntry | undefined> {
```
a-frazier: **Scene-state sourcing is now split across two implementations.**

`readSceneSource` here sources `sceneEntities` / `currentLocationId` / `sceneMetadata` from the last `ai_reply` / `opening` row in SQLite. But two other paths for the same turn still derive scene from the entriesStore reader-window tail via `inheritedEntryMetadata(entries.at(-1)?.metadata)`:

- `retrievalPhase` — `lib/pipeline/definitions/per-turn-retrieval.ts:62`
- `narrativePhase`'s `buildPiggybackActions` `previousMetadata` — `lib/pipeline/definitions/per-turn.ts:177-178`

The tail is usually this turn's `user_action`, so the two agree only while `user_action` scene-metadata inheritance is perfect. If a `user_action` fails to inherit, or the planned scene-field editor stamps scene state onto a `user_action` row, retrieval ranks against one scene (or an empty one) while the prompt shows a different scene for the same turn. Pre-refactor both paths read the same tail and could not disagree.

Worth a deliberate decision: either route retrieval/apply through the same AI-row read, or document why the tail is acceptable for those two.

lib/pipeline/definitions/per-turn-piggyback.ts
```diff
@@ -144,22 +143,20 @@ export async function* piggybackFallbackClassifierPhase(
...
-    entries: previousEntry ? [previousEntry, tail] : [tail],
-    entities,
-    definition: open.definition,
-    settings: open.settings,
-    idMap,
+  // story-related prompt like any other, not a special case. Its template
+  // renders `lastTurns` rather than `entries`: the user's action can itself
+  // carry state changes ("I put the sword away"), not just the AI's reply, and
+  // that pair is a fixed contract the story's buffer knobs must not cut.
+  const load = await buildGenerationContext(ctx, {
```
a-frazier, line 155: **Fallback classifier reads its turn pair from SQLite but writes the extracted metadata to a store-derived id.**

The prompt's turn pair now comes from `readLastTurns(ctx.db, branchId)` (via `buildGenerationContext` -> `lastTurns`). But the metadata delta still targets `tail.id` where `tail = working.set.entries.at(-1)` (line 133), and `previousMetadata` is computed from `previousEntry = entries.at(-2)` (line 135, used at line 203) — both off the entriesStore working set.

Before this PR the prompt entries and the write target were the same array (`previousEntry ? [previousEntry, tail] : [tail]`). If `entriesStore` lags SQLite for the branch tail (row written outside the store, reader window not yet refreshed, a concurrent branch op), the classifier reasons over one entry pair and writes the result to a different entry, and `previousMetadata` is derived from a row the prompt never showed.

Consider deriving the write target and `previousMetadata` from the same `readLastTurns` result the prompt used.

lib/prompts/engine.ts
```diff
@@ -22,6 +22,29 @@ export function createEngine(pack: Pack): Liquid {
...
+  let perTemplate = globalsByEngine.get(engine)
+  if (!perTemplate) {
+    perTemplate = new Map()
+    globalsByEngine.set(engine, perTemplate)
+  }
+
+  const cached = perTemplate.get(templateId)
+  if (cached) return cached
+
+  const globals = new Set(engine.globalVariablesSync(engine.parseFileSync(templateId)))
```
a-frazier, line 43: **The read-skipping optimization leans on include-following that nothing tests.**

`buildGenerationContext` skips `readPromptBuffer` / `readLastTurns` / `readSceneSource` unless the template is reported to read the variable, and that check bottoms out in `engine.globalVariablesSync(engine.parseFileSync(templateId))` here — which must resolve `{% include %}`d macros for the gating to be correct.

All three bundled templates reference their gated vars (`entries`, `lastTurns`, scene) directly, so there is no current failure and no test exercises the macro-only path. If someone later moves e.g. `{% for entry in entries %}` into a shared macro, or a custom pack reads `entries` only via an include that sync partial analysis doesn't follow, the template renders against an empty buffer with no error — a silently truncated prompt.

A fixture template that reads a gated var only through an `{% include %}` would lock this down.

lib/pipeline/definitions/generation-context.ts
```diff
@@ -1,54 +1,56 @@
...
-  // a truncation seam, not a per-kind query. What reaches a template is
-  // composePromptBuffer's window over this list, so a caller handing over a
-  // deliberate short tail can still have it cut — or emptied — by the story's
-  // own buffer knobs, which the caller does not control.
-  entries: readonly StoryEntry[]
-  entities: readonly Entity[]
-  definition: StoryDefinition
-  settings: StorySettings
-  idMap: IdBiMap
+import { RETRIEVAL_INTERMEDIATE_KEY } from './per-turn-retrieval'
```
a-frazier: **This import pulls the whole retrieval/embedder phase graph into every context builder.**

`readRetrievalOutcome` moved into this module, which is imported by all three generation phases and their tests. The only symbol actually needed from `./per-turn-retrieval` is the string literal `RETRIEVAL_INTERMEDIATE_KEY` (`'retrieval'`), but importing it drags in `runRetrieval`, `@/lib/embedder`, `@/lib/embedder-swap`, and `@/lib/probe` transitively — widening module-load cost and setting up a latent import cycle the moment retrieval ever needs the context builder.

Moving the constant to `../types` (or a leaf module) keeps the dependency edge one-directional.

a-frazier (commented):
Automated code-review pass (high effort) on the `buildGenerationContext` -> async data-source refactor. No crash-class bug found; the faithful buffer-logic port is well covered by the `promptBufferTake` / `readPromptBuffer` suites and the new "data source" describe block. The four inline notes below are consistency seams and latent fragilities rather than defects in the current build. Posting for author triage — none are hard blockers.

PR #484, opened 2026-09-05
Report what a turn is doing while it generates

collegatore:
## Why

The wait before the first narration token could run to forty seconds behind a three-dot animation. A single turn can make two tier-3 selections, a chapter query per agentic step, and up to three narrative retries — none of it visible. Debug Mode answers a different question (what was sent and received) through a modal or a second window, and is uncomfortable to leave running while reading.

## What this does

A turn records its work as a tree of timed steps.

- **While generating**, the story view replaces the ellipsis with the **deepest step currently running** — `query ch.12 · 6.2s` rather than "Retrieving" — plus the turn's elapsed time, expandable to the full timeline.
- **After the turn**, the record stays reachable from the finished entry for the **last five turns** of the session, so a turn that felt slow can be examined afterwards.
- **LLM-backed steps are marked**, so the expensive work is distinguishable from the free work at a glance.
- A new **three-state interface setting** (`Off` / `Status line` / `Full timeline`) controls it, independent of Debug Mode.

Newly visible: world state and lorebook tier-3 selection, each agentic step and tool call against its budget, each chapter query, timeline fill, narrative prompt assembly and time-to-first-token, and **narrative empty-response retries** — which were previously invisible even in Debug Mode.

## Approach

- `services/activity/` holds the record and the reading views (nesting, deepest running step, durations, retention), all plain TypeScript. `stores/activity.svelte.ts` is the reactive shell over it, following the `debug.svelte.ts` pattern.
- Phases reach an injected `ActivityReporter` on `PipelineDependencies` rather than importing the store — the rule `docs/architecture/overview.md` names as what keeps them testable without a provider. `NO_ACTIVITY` stands in where none is wired, so existing phase tests needed no changes.
- Services under `services/ai/` write to the store directly, following the debug store's precedent; their tests mock it the same way.
- Nesting is by explicit parent id threaded through existing options objects. An implicit stack would not survive Stage A's two concurrent branches or the parallel post-narrative phases.
- Retrieval reports through the `RetrievalEvent` record it already keeps; chapter queries reuse the `durationMs` the query budget already measured.
- Phases with several completion paths are wrapped by `trackPhase` rather than instrumented one exit at a time — `ImagePhase` alone has five.

## Reporting does not alter the turn

No prompt, request count, or retry policy changes. The narrative retry loop is reported but untouched, with a test asserting the attempt count and outcome are unchanged. Every store write is guarded and the display sits inside a `<svelte:boundary>`, so a fault in the record cannot take down the generation it describes. Records are session-only: nothing is persisted, exported, or carried across a restart. No migration.

## Verification

`npm run check` 0 errors · `npm test` 1425 passed · `npm run lint` 0 errors · `vite build` succeeds.

**Not verified:** the in-app checks. There is no DOM in the test environment, so the Svelte half is unexercised by the suite, and I could not run the Tauri app here. Worth a look before merge:

- the collapsed line tracking through a long retrieval, and how the tie-break between the two concurrent Stage A branches reads
- Stop mid-turn, then opening that turn's record
- toggling the setting mid-turn (the in-flight record is discarded by design)
- the mobile path — the badge is desktop-only like the token badge beside it, with an overflow-menu equivalent

Planning artifacts live in the `aventuras-specs` store under `generation-activity-timeline`.

## Summary by CodeRabbit

- **New Features**
  - Added optional Generation Activity reporting during AI responses.
  - Choose Off, Status line, or Full timeline in Interface settings.
  - View live progress, nested steps, statuses, and elapsed durations for generation and retrieval tasks.
  - Activity details appear in story entries and streaming responses without interrupting narration.

- **Bug Fixes**
  - Activity tracking now closes reliably after successful, interrupted, or failed operations.
  - Display errors no longer hide story content.
  - Improved cleanup when generation streams stop unexpectedly.

- **Tests**
  - Added coverage for activity tracking, timelines, durations, retention, cleanup, and generation workflows.

### DIFF
--- a/src/lib/components/story/ActionInput.svelte
+++ b/src/lib/components/story/ActionInput.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import { tick } from 'svelte'
   import { ui, type RetrievalCacheKey } from '$lib/stores/ui.svelte'
+  import { activity } from '$lib/stores/activity.svelte'
   import { toRetrievalSnapshot } from '$lib/services/ai/retrieval'
   import { buildTimelineFillBlock } from '$lib/services/ai/generation'
   import { joinPromptBlocks } from '$lib/utils/promptBlocks'
@@ -236,6 +237,7 @@
    */
   function buildPipelineDependencies(storyId: string): PipelineDependencies {
     return {
+      activity,
       shouldUseAgenticRetrieval: () =>
         aiService.shouldUseAgenticRetrieval(settings.systemServicesSettings.timelineFill),
       runAgenticRetrieval: (options) =>
@@ -248,7 +250,7 @@
       // The chapter-read budget is derived from this story's own chapterization threshold, so
       // it is bound here with the rest of the store rather than read from settings: a chapter
       // is about `tokenThreshold` tokens by construction. See `story.chapterReadBudget`.
-      runTimelineFill: (visibleEntries, chapters, alreadyInContext) =>
+      runTimelineFill: (visibleEntries, chapters, alreadyInContext, activityParentId) =>
         aiService.runTimelineFill(
           storyId,
           visibleEntries,
@@ -256,6 +258,7 @@
           story.getChapterEntries.bind(story),
           alreadyInContext,
           story.chapterReadBudget,
+          activityParentId,
         ),
       answerChapterQuestion: (chapterNumber, question, chapters) =>
         aiService.answerChapterQuestion(
@@ -480,6 +483,7 @@
     }
 
     ui.setGenerating(true)
+    activity.startTurn(narrationEntryId)
     ui.clearGenerationError()
     ui.clearActionChoices(story.currentStory.id)
     ui.startStreaming(visualProseMode, streamingEntryId)
@@ -826,6 +830,8 @@
       ui.endStreaming()
       ui.setGenerating(false)
       ui.setGenerationStatus('')
+      // Closes the turn even when a step was left running, so the record is bounded.
+      activity.endTurn()
       activeAbortController = null
 
       // Android: always stop the foreground service when generation ends
--- a/src/lib/components/story/ActionInput.svelte
+++ b/src/lib/components/story/ActionInput.svelte
@@ -504,16 +504,6 @@
     }
 
     ui.setGenerating(true)
-    const inputTranslation = options?.inputTranslation
-    activity.startTurn(narrationEntryId, inputTranslation?.startedAt)
-    if (inputTranslation) {
-      activity.recordStep('Translating input', {
-        isLLM: true,
-        startedAt: inputTranslation.startedAt,
-        durationMs: inputTranslation.durationMs,
-        status: inputTranslation.failed ? 'failed' : 'done',
-      })
-    }
     ui.clearGenerationError()
     ui.clearActionChoices(story.currentStory.id)
     ui.startStreaming(visualProseMode, streamingEntryId)
@@ -541,6 +531,19 @@
     ui.resetBackgroundedFlag()
 
     try {
+      // Inside the try: only its `finally` closes the turn, and a throw before that point
+      // would leave a record nothing can close.
+      const inputTranslation = options?.inputTranslation
+      activity.startTurn(narrationEntryId, inputTranslation?.startedAt)
+      if (inputTranslation) {
+        activity.recordStep('Translating input', {
+          isLLM: true,
+          startedAt: inputTranslation.startedAt,
+          durationMs: inputTranslation.durationMs,
+          status: inputTranslation.failed ? 'failed' : 'done',
+        })
+      }
+
       const worldState = story.worldStateSnapshot
 
       const storyPosition = story.entries.length
--- a/src/lib/components/story/ActionInput.svelte
+++ b/src/lib/components/story/ActionInput.svelte
@@ -67,13 +67,29 @@
   // Translation Helper
   // ============================================================================
 
+  /** What the input translation cost, for the turn record. See `InputTranslationTiming`. */
+  type InputTranslationTiming = { startedAt: number; durationMs: number; failed: boolean }
+
   async function translateUserInput(
     content: string,
     translationSettings: typeof settings.translationSettings,
-  ): Promise<{ promptContent: string; originalInput: string | undefined }> {
+  ): Promise<{
+    promptContent: string
+    originalInput: string | undefined
+    timing?: InputTranslationTiming
+  }> {
     if (!TranslationService.shouldTranslateInput(translationSettings)) {
       return { promptContent: content, originalInput: undefined }
     }
+
+    // Measured here because this runs before the generation path opens the turn record, and
+    // it is a model call on the same critical path as everything the record does cover.
+    const startedAt = Date.now()
+    const timing = (failed: boolean): InputTranslationTiming => ({
+      startedAt,
+      durationMs: Date.now() - startedAt,
+      failed,
+    })
 
     try {
       log('Translating user input', {
@@ -88,10 +104,14 @@
         originalLength: content.length,
         translatedLength: result.translatedContent.length,
       })
-      return { promptContent: result.translatedContent, originalInput: content }
+      return {
+        promptContent: result.translatedContent,
+        originalInput: content,
+        timing: timing(false),
+      }
     } catch (error) {
       log('Input translation failed (non-fatal), using original', error)
-      return { promptContent: content, originalInput: undefined }
+      return { promptContent: content, originalInput: undefined, timing: timing(true) }
     }
   }
 
@@ -454,6 +474,7 @@
       countStyleReview?: boolean
       styleReviewSource?: string
       cachedRetrievalResult?: RetrievalResult | null
+      inputTranslation?: InputTranslationTiming
     },
   ) {
     const countStyleReview = options?.countStyleReview ?? true
@@ -483,7 +504,16 @@
     }
 
     ui.setGenerating(true)
-    activity.startTurn(narrationEntryId)
+    const inputTranslation = options?.inputTranslation
+    activity.startTurn(narrationEntryId, inputTranslation?.startedAt)
+    if (inputTranslation) {
+      activity.recordStep('Translating input', {
+        isLLM: true,
+        startedAt: inputTranslation.startedAt,
+        durationMs: inputTranslation.durationMs,
+        status: inputTranslation.failed ? 'failed' : 'done',
+      })
+    }
     ui.clearGenerationError()
     ui.clearActionChoices(story.currentStory.id)
     ui.startStreaming(visualProseMode, streamingEntryId)
@@ -1006,10 +1036,11 @@
       story.currentStory.timeTracker,
     )
 
-    const { promptContent, originalInput } = await translateUserInput(
-      content,
-      settings.translationSettings,
-    )
+    const {
+      promptContent,
+      originalInput,
+      timing: inputTranslation,
+    } = await translateUserInput(content, settings.translationSettings)
 
     const userActionEntry = await story.addEntry('user_action', promptContent)
 
@@ -1025,7 +1056,7 @@
     // entry the narrator reads holds `promptContent`. Passing the raw text here left
     // retrieval and classification working from a different wording than the narration —
     // and the retry path already passes `promptContent`, so the two disagreed.
-    await generateResponse(userActionEntry.id, promptContent)
+    await generateResponse(userActionEntry.id, promptContent, { inputTranslation })
   }
 
   async function handleStopGeneration() {
@@ -1201,10 +1232,11 @@
 
     await tick()
 
-    const { promptContent, originalInput } = await translateUserInput(
-      backup.userActionContent,
-      settings.translationSettings,
-    )
+    const {
+      promptContent,
+      originalInput,
+      timing: inputTranslation,
+    } = await translateUserInput(backup.userActionContent, settings.translationSettings)
     const userActionEntry = await story.addEntry('user_action', promptContent)
 
     if (originalInput) {
@@ -1225,6 +1257,7 @@
         countStyleReview: false,
         styleReviewSource: 'retry-last-message',
         cachedRetrievalResult: cacheKey ? ui.retrievalResultFor(cacheKey) : null,
+        inputTranslation,
       })
     } finally {
       ui.setRetryingLastMessage(false)
--- a/src/lib/components/story/StoryEntry.svelte
+++ b/src/lib/components/story/StoryEntry.svelte
@@ -19,6 +19,7 @@
     Volume2,
     Image as ImageIcon,
     Copy,
+    Clock,
     MoreVertical,
   } from '@lucide/svelte'
   import { aiService } from '$lib/services/ai'
@@ -52,6 +53,9 @@
   import { database } from '$lib/services/database'
   import { onMount } from 'svelte'
   import ReasoningBlock from './ReasoningBlock.svelte'
+  import ActivityStatus from './ActivityStatus.svelte'
+  import { activity } from '$lib/stores/activity.svelte'
+  import { formatDuration, turnDuration } from '$lib/services/activity'
   import { countTokens } from '$lib/services/tokenizer'
   import { errMessage } from '$lib/utils/error'
   import { Button } from '$lib/components/ui/button'
@@ -115,6 +119,15 @@
 
   // Generation info shown in the "info" popover (model/profile/effort/timestamp)
   const showInfo = $derived(entry.type === 'narration')
+
+  // Only while this turn's record is still retained; an evicted one offers nothing rather
+  // than an empty panel. See RETAINED_TURNS.
+  const activityRecord = $derived(
+    settings.uiSettings.activityReporting !== 'off' && entry.type === 'narration'
+      ? activity.recordFor(entry.id)
+      : null,
+  )
+  let showActivityRecord = $state(false)
 
   function formatStoryTime(time: TimeTracker | null | undefined): string {
     if (!time) return ''
@@ -1356,6 +1369,20 @@
       <span class="text-muted-foreground ml-0.5">tokens</span>
     </span>
 
+    <!-- How long the turn took, opening its timeline. Hidden on narrow screens for the same
+         reason as the token badge above; the overflow menu carries it there instead. -->
+    {#if activityRecord}
+      <button
+        type="button"
+        class="bg-muted text-muted-foreground hover:text-foreground hidden rounded px-1.5 py-0.5 text-[11px] tabular-nums transition-colors sm:inline"
+        aria-expanded={showActivityRecord}
+        title="What this turn spent its time on"
+        onclick={() => (showActivityRecord = !showActivityRecord)}
+      >
+        {formatDuration(turnDuration(activityRecord, activityRecord.endedAt ?? Date.now()))}
+      </button>
+    {/if}
+
     <!-- Spacer to push buttons to the right -->
     <div class="flex-1"></div>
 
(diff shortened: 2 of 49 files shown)


src/lib/services/ai/retrieval/TimelineFillService.ts
```diff
@@ -370,7 +374,12 @@ export class TimelineFillService extends BaseAIService {
...
 
+    const planStepId = activity.startStep('Planning questions', {
+      parentId: activityParentId,
+      isLLM: true,
+    })
     const queries = await this.generateQueries(storyId, visibleEntries, chapters, alreadyInContext)
+    activity.endStep(planStepId, 'done', `${queries.length} questions`)
     if (queries.length === 0) {
       return { queries: [], responses: [] }
     }
```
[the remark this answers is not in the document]
collegatore: Fixed in 402928ed — both `Planning questions` and the per-group `Reading ch.N` step now close as `failed` (or `skipped` on an abort) before rethrowing, instead of surviving to `endTurn` and being swept up as `interrupted`; two regression tests cover the render-rejection and abort paths.

src/lib/services/ai/retrieval/AgenticRetrievalService.ts
```diff
@@ -28,6 +28,8 @@ import {
 } from '../sdk/tools'
 import { ContextBuilder } from '$lib/services/context'
 import { debug } from '$lib/stores/debug.svelte'
+import { activity } from '$lib/stores/activity.svelte'
```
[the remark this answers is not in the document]
collegatore: Not changing this: the convention cited says the opposite. `docs/development/testing.md` states rune modules cannot be *imported* by tests and that "services reach the stores through `vi.mock('$lib/stores/…')`" — which is what these three do, exactly as they already did for the `debug` store before this PR. The generation phases, which the architecture docs do protect, stay store-free via the injected `ActivityReporter`. Threading a reporter through three more service APIs is considered and rejected in this change's design.md, since it buys no behavioural change and would make activity inconsistent with `debug` and `settings`.

src/lib/services/ai/retrieval/tier3Selection.test.ts
```diff
@@ -39,6 +47,7 @@ import {
   type Tier3SelectionResult,
 } from './tier3Selection'
 import { TIER3_SELECTION_CACHE_POSITIONS } from '../core/defaults'
+import { activity } from '$lib/stores/activity.svelte'
```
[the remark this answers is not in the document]
collegatore: Not changing this: `vi.mock` is hoisted above the import, so the specifier resolves to the mock factory and the real rune module is never loaded — the suite passes, which is the proof. The guideline forbids importing a rune module, and prescribes `vi.mock` as the way services reach stores. There is existing precedent for the same pattern in `src/lib/services/ai/image/providers/registry.test.ts`, which imports `settings` from `$lib/stores/settings.svelte` after mocking it.

src/lib/services/ai/retrieval/AgenticRetrievalService.ts
```diff
@@ -333,6 +359,12 @@ export class AgenticRetrievalService extends BaseAIService {
       log('Agent run failed -- salvaging what it gathered', { failure, steps: stepsTaken })
     }
 
+    activity.endStep(
```
a-frazier: **Severity: low.**

On a mid-retrieval abort the `Agent` step is never closed before the error is rethrown. When the user stops generation, `agent.generate()` throws and `if (signal?.aborted) throw error` (≈ line 357) re-raises *before* this `endStep`. The step stays `running`; `RetrievalPhase`'s catch only closes the parent memory step, so `endTurn()` later sweeps `Agent` as skipped / `interrupted` rather than closed cleanly.

Same pattern that was fixed in `TimelineFillService` (402928ed). A `try { … } finally { activity.endStep(agentStepId, …) }` around the agent run, or an explicit close on the abort path, would keep this consistent.

src/lib/stores/activity.svelte.ts
```diff
@@ -0,0 +1,118 @@
...
+  /** The turn in flight. Touches `version` so callers re-read as it grows. */
+  get activeTurn(): ActivityTurn | null {
+    void this.version
+    return this.recorder.activeTurn
+  }
+
+  /** The retained record for an entry, or null once evicted. */
+  recordFor(entryId: string): ActivityTurn | null {
+    void this.version
+    return findTurnByEntryId(this.recorder.snapshot(), entryId)
```
a-frazier: **Severity: low (performance, generation path).**

`recordFor()` / `hasRecord()` call `recorder.snapshot()`, which spreads every retained turn and all of its steps, on every invocation. This runs from a `$derived` in each mounted narration `StoryEntry` and re-fires on every `version` bump — dozens to hundreds of times per agentic turn — including for the entries whose record was already evicted. The fresh-copy identity churn also forces `ActivityTimeline`'s `buildTree` and `ActivityStatus`'s `$effect` to re-run each time.

A per-`version` memoized snapshot, or a non-copying `findTurnByEntryId` over the live turns for these two lookups, avoids the repeated allocation while generation is running.

src/lib/services/generation/GenerationPipeline.ts
```diff
@@ -166,26 +184,32 @@ export class GenerationPipeline {
...
-          activeThreads: cfg.activeThreads,
-          lorebookEntries: ctx.worldState.lorebookEntries,
-          promptContext: cfg.promptContext,
-          worldState: ctx.worldState,
-          narrativeResponse: r.narrative.content,
-          pov: cfg.pov,
-          translationSettings: cfg.translationSettings,
-          abortSignal: ctx.abortSignal,
-        }),
+        background: this.tracked(
```
a-frazier: **Severity: low.**

When one merged phase here throws (rather than yielding a fatal error event), `mergeGenerators`' `Promise.race` rejects and the loop exits without `.return()`-ing the sibling generators. Their `trackPhase` `finally` (`activity.endStep`) never runs, so the surviving steps — e.g. `Background image` / `Suggestions`, and the `Classification` / `Translation` pair in the `imageDeps` merge below — leak as `running` until `endTurn` sweeps them as `interrupted`.

This is pre-existing `mergeGenerators` behaviour that the new instrumentation now makes visible. Worth deciding whether `mergeGenerators` should `.return()` its outstanding siblings on rejection.

src/lib/components/story/ActionInput.svelte
```diff
@@ -480,6 +483,7 @@
     }
 
     ui.setGenerating(true)
+    activity.startTurn(narrationEntryId)
```
a-frazier: **Severity: low.**

`activity.startTurn()` runs before the `try` whose `finally` calls `activity.endTurn()` (≈ line 830). The statements in between — `ui.resetBackgroundedFlag()`, `new InlineImageTracker(...)`, the `window.AndroidBridge` calls — would leave the turn open if any of them throws: `recorder.current` stays set until the next `startTurn` replaces it and it is eventually evicted.

Moving `startTurn` inside the `try` closes the window. (`startTurn` is guarded so it can't throw itself — this is only about the unpaired region.)

src/lib/stores/settings.svelte.ts
```diff
@@ -3107,6 +3126,9 @@ class SettingsStore {
     // Reset UI settings
     this.uiSettings = getDefaultUISettings()
     await this.setNavPanelWidth(this.uiSettings.navPanelWidth)
+    // Through the setter: the defaults object alone leaves the live recorder on its old mode
+    // and never writes the key, so the previous mode came back on the next start.
```
a-frazier: **Severity: trivial (style).**

This comment narrates the prior buggy behaviour ("the previous mode came back on the next start"), which the repo's `CLAUDE.md` Comments section explicitly rules out (*"Do not write: history — 'this was a bug where…'"*).

State the rule instead, e.g.: `// Route through the setter so the live recorder and the persisted key are updated, not just the in-memory settings object.`

a-frazier (commented):
Automated review pass (high effort). The prior CodeRabbit findings — and the two already fixed in `abaf24cf` (Reset All) and `402928ed` (prompt-render failure closing steps) — were checked and are not repeated here.

Five new observations below, all minor / low severity. The PR is well-guarded overall: every store write wrapped, display inside `<svelte:boundary>`, `NO_ACTIVITY` fallback.

| # | File | Severity |
|---|------|----------|
| 1 | `AgenticRetrievalService.ts` | low — `Agent` step not closed on mid-retrieval abort |
| 2 | `activity.svelte.ts` | low (perf) — `recordFor()` deep-copies all turns on every call |
| 3 | `GenerationPipeline.ts` | low — merged phases leak on sibling throw (pre-existing `mergeGenerators`) |
| 4 | `ActionInput.svelte` | low — `startTurn()` outside the `try` that ends it |
| 5 | `settings.svelte.ts` | trivial — comment narrates a past bug (CLAUDE.md) |

a-frazier (approved):
Fixing the feedback is optional -- it's all pretty minor. If you'd rather do it in a separate PR, let me know and I'll merge this one.

src/lib/services/ai/retrieval/AgenticRetrievalService.ts
collegatore: Fixed in e93c0fc8 — the abort path now closes both the iteration in flight and the `Agent` step as `skipped` (carrying the step count) before rethrowing, matching the TimelineFillService fix.

src/lib/stores/activity.svelte.ts
collegatore: Fixed in e93c0fc8 — added `ActivityRecorder.find()`, which reads the live turn, and `recordFor` uses it instead of `snapshot()`. Worth noting the identity churn was more than allocation: a new object per `version` bump also re-ran `ActivityStatus`'s `$effect`, tearing down and recreating the 1s ticker on a finished record.

src/lib/services/generation/GenerationPipeline.ts
collegatore: Fixed in e93c0fc8 — decided in favour of cleaning up. `mergeGenerators` now closes any outstanding siblings in a `finally`, so it covers the consumer walking away as well as a rejection, with `allSettled` so a sibling failing to close cannot mask the original error. Three tests added; the pipeline is its only caller.

src/lib/components/story/ActionInput.svelte
collegatore: Fixed in e93c0fc8 — `startTurn` and the `Translating input` record moved inside the `try`, so the only path that opens a turn is the one whose `finally` ends it.

src/lib/stores/settings.svelte.ts
collegatore: Fixed in e93c0fc8 — now states the rule: routing through the setter is what reaches the live recorder and the persisted key. I swept the rest of the diff for the same thing and rewrote one more, in `AgenticRetrievalService`, that leaned on a measurement.

src/lib/services/activity/duration.test.ts
```diff
@@ -0,0 +1,78 @@
+import { describe, it, expect } from 'vitest'
+import { formatDuration, formatStepDuration, stepDuration, turnDuration } from './duration'
```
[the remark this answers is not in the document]
collegatore: Not changing this: the rule permits it. Its own message reads "Import from the service public API (index.ts), not internal modules. **Use relative imports within the same service**" — and `duration.test.ts` is inside `services/activity/`, the same service as `duration.ts`. `npx eslint src/lib/services/activity` reports zero errors and zero warnings, and every sibling test in the module imports the same way (`./tree`, `./recorder`, `./retention`, `./trackPhase`). Switching just this one would make it the odd file out.

src/lib/utils/async.ts
```diff
@@ -63,17 +63,26 @@ export async function* mergeGenerators<
...
+        getNext(key)
+      }
     }
+  } finally {
+    // One generator throwing, or the consumer walking away, leaves the rest mid-iteration.
+    // They are owned here, so their cleanup is closed here rather than left to collection --
+    // `allSettled` because a sibling failing to close must not mask the original failure.
+    await Promise.allSettled(
+      Array.from(activeGenerators.values()).map((gen) => gen.return(undefined)),
+    )
```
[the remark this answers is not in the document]
collegatore: Confirmed and fixed in dd03440e. Cleanup is started but no longer awaited — each sibling is still closed, so its own `finally` runs once its read settles, but the failure reaches the caller immediately.

Worth recording that this was stronger than a delay: with every sibling parked on an unanswered read, `allSettled` never settled and `mergeGenerators` did not return at all. The regression test (`propagates the failure without waiting on a sibling read that has not answered`) hangs and times out against the awaiting form — verified by reverting to `await` and watching it fail at 3009ms.

Trade-off taken deliberately: a step whose `endStep` now lands after `endTurn` is swept as `interrupted` rather than carrying its own status. That is cosmetic and confined to a path where a phase throws rather than yielding an error event; blocking error propagation on an uncancellable model call is not.

src/lib/utils/async.test.ts
```diff
@@ -44,3 +44,110 @@ describe('pLimit', () => {
...
+      })(),
+    })
+
+    await expect(
+      (async () => {
+        for await (const _ of merged) void _
+      })(),
+    ).rejects.toThrow('boom')
+
+    releaseSibling?.()
```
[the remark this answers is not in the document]
collegatore: Valid — fixed in ec4b9099, and I went further than the proposed guard. The sibling reached its unresolved read only because the scheduler got there first; the ordering was never enforced, so on a different interleaving the test would have passed against the awaited form it exists to reject.

The throw is now held until the sibling signals it is parked, making the pending read a precondition rather than a race, and the release is asserted and called unconditionally as suggested. Re-verified by reverting to `await`: it fails at 3015ms, same as before.

src/lib/components/story/StreamingEntry.svelte
```diff
@@ -103,6 +127,17 @@
     </div>
   </div>
 
+  <!-- Activity report. Above the content on purpose: it keeps one place on screen while the
+       narration grows underneath it, rather than being pushed off the bottom.
+       A bystander to the entry -- a fault rendering it must not take the narration with it. -->
+  <svelte:boundary>
```
a-frazier: This `<svelte:boundary>` has no `failed` snippet, `onerror`, or `pending` prop, so per Svelte 5 semantics it is inert — an error thrown while rendering `ActivityStatus` / `ActivityTimeline` propagates straight past it to the next boundary up, exactly as if this wrapper weren't here. The comment's promise ("a fault rendering it must not take the narration with it") isn't delivered: the streaming narration entry unmounts with it.

Plausible throw sources that aren't network/concurrency (so the single-user local setup doesn't rule them out): a `@lucide/svelte` version bump changing the icon component API, an icon import resolving to `undefined`; a `$derived` throwing on unexpected step shape (`startedAt` NaN/string, a field a later recorder change adds); a `parentId` cycle in `buildTree`/`flattenTree`, which — unlike `deepestRunningStep`/`rootStep` — has no cycle guard, giving unbounded recursion → `RangeError`. None caught by `check`/`lint`/tests (no DOM in the test env).

Minimal fix: `onerror={(e) => console.warn('[activity]', e)}` plus an empty `{#snippet failed()}{/snippet}`.

src/lib/components/story/StoryEntry.svelte
```diff
@@ -1775,6 +1815,16 @@
         />
       {/if}
 
+      <!-- The report is a bystander to the entry: a fault rendering it must not
+         take the narration with it. -->
+      <svelte:boundary>
```
a-frazier: Same inert `<svelte:boundary>` as in `StreamingEntry.svelte` — no `failed` / `onerror` / `pending`, so it catches nothing and the comment above it ("a fault rendering it must not take the narration with it") doesn't hold.

This one is the worse of the two: `activityRecord` comes from `activity.recordFor(entry.id)` over a retained-turn record that outlives the turn. If that record's shape ever lags a recorder change (the local equivalent of migration skew for the retained-turns data), the malformed `turn` flows into `<ActivityStatus>` when a user clicks "Show generation activity" on a past entry, and an uncaught throw here tears down the whole `StoryEntry` — including the narration text the user is actually reading — not just the activity panel.

Minimal fix: `onerror={(e) => console.warn('[activity]', e)}` plus an empty `{#snippet failed()}{/snippet}`.

src/lib/services/activity/recorder.ts
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * Nesting is by explicit parent id rather than an implicit stack: Stage A's two branches and
+ * the post-narrative phases run concurrently, so there is no single "current" step to push
+ * onto.
+ */
+
+import { findTurnByEntryId, retainTurns, RETAINED_TURNS } from './retention'
+import type { ActivityStatus, ActivityStep, ActivityTurn } from './types'
+
+/** How much of a turn's activity the story view reports. See design.md. */
```
a-frazier: "See design.md" — no `design.md` exists in the repo; this PR put the rationale in `docs/architecture/overview.md`. Per CLAUDE.md ("docs/ holds the architecture, one file per area"), this and the matching reference in `retention.ts` should point at `docs/architecture/overview.md` (the activity-reporting section) so the next reader isn't sent looking for a file that was never added.

src/lib/services/activity/retention.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Activity Retention
+ *
+ * Records are kept for the current session only, bounded by a count of turns. See
+ * design.md — Retention is five turns.
```
a-frazier: Same dangling reference as in `recorder.ts:15` — "See design.md" points at a file that doesn't exist. The retention rationale ("five turns") lives in `docs/architecture/overview.md` after this PR; point there instead.

a-frazier (commented):
Three findings from a code review that still stand against the current head (`a7f265cc`). Findings on `formatDuration` rounding and the `ActivityTimeline` `depth` prop from the same review are already fixed in this branch.

- **`<svelte:boundary>` is inert in two places** (`StreamingEntry.svelte`, `StoryEntry.svelte`) — no `failed`/`onerror`/`pending`, so the fault-isolation the comments describe isn't wired. Details inline.
- **Dangling `design.md` references** in `recorder.ts` and `retention.ts` — should point at `docs/architecture/overview.md`.

src/lib/components/story/StreamingEntry.svelte
collegatore: Confirmed and fixed in 7709fa16. Verified against the runtime rather than the docs — `svelte/src/internal/client/dom/blocks/boundary.js` opens its `error` with `if (!this.#props.onerror && !this.#props.failed) throw error`, so the wrapper was doing nothing at all.

Added `onerror` only, no `failed` snippet: `#handle_error` destroys the main effect and renders nothing when there is no fallback, which is the wanted outcome — the report disappears, the narration stays.

On the `RangeError` route specifically: it is not reachable. A step in a `parentId` cycle always has a resolvable parent, so `buildTree` never makes it a root, and `flattenTree` only descends from roots — a cycle ends up detached and simply unrendered rather than recursed into. The other sources you list stand, and the boundary covers them.

src/lib/components/story/StoryEntry.svelte
collegatore: Fixed in 7709fa16, same `onerror` treatment. Agreed this is the worse of the two: the record outlives the turn, so a shape mismatch surfaces on a click on an old entry, and an uncaught throw there would tear down the narration the reader came for.

src/lib/services/activity/recorder.ts
collegatore: Fixed in 7709fa16 — points at `docs/architecture/overview.md` now. The `design.md` is real but lives in the `aventuras-specs` store, so it was unreachable from the repo.

src/lib/services/activity/retention.ts
collegatore: Fixed in 7709fa16 — points at the Activity reporting section of `docs/architecture/overview.md`, which carries the `RETAINED_TURNS` bound.

a-frazier (approved):

PR #489, opened 2026-09-06
Deal with frequent Android freezes

collegatore:
## The bug

On Android the app regularly reached a state where it rendered correctly but ignored every touch — taps, scrolling and the action input all dead, while generation kept running and entries kept streaming in. Only the floating scroll buttons responded, and the state survived navigating away from the story view, persisting for the rest of the session.

That one surviving control is the whole diagnosis. `StoryView.svelte:583` is the only element in the story view that sets `pointer-events: auto`, and `pointer-events` is inherited — so `pointer-events: none` on `<body>` kills the entire tree *except* anything that re-declares it. Not an overlay covering the app; the input was severed at the root.

Both modal libraries write `pointer-events: none; overflow: hidden` to `document.body` while an overlay is open and restore it on the teardown paths they control. An overlay unmounted while still open never runs that restore, and the lock is permanent.

## Why the existing net didn't catch it

`releaseOrphanScrollLock` and the `AppShell` watcher already existed for exactly this. The watcher is triggered by a `MutationObserver` on `body.style` — and an orphaned lock is by definition the case where nothing writes to `body.style` again. It gets one look, at the moment the overlay is still legitimately open, and is never called back.

This is also why it was Android-only: the only other code that writes `body.style` is the desktop pane-resize handler, so on desktop the next splitter drag silently rescued the session.

## What changed

**Recovery is driven by real input** (`AppShell.svelte`). Capture-phase `pointerdown`, `keydown` and `wheel` listeners on `document`, sharing one release with the existing watcher, which is kept. The correctness argument mentions no library: if a real input event reaches the document while the body is locked and no owner is present, the lock is orphaned. A locked page still delivers those events — `<body>` leaves hit testing but `<html>` keeps `pointer-events: auto`, so the event targets `documentElement`. `wheel` is included because `overflow: hidden` makes scrolling the first thing a user tries, and a wheel is not a `pointerdown`.

**Ownership detection corrected** (`scrollLock.ts`). Two defects found in review, both of which would have survived a fix to the trigger alone:

- Entitlement tested `data-state="open"`, but a lock owner is mounted with its *element* and still holds the lock while animating closed. Releasing in that window could leave a modal opened behind a still-exiting one displayed with no lock at all. Now matches presence.
- Presence alone is too broad: `[data-state]` is what separates a library-managed owner from a hand-rolled overlay carrying the same ARIA role. The app has two non-owners (`CharacterPanel` expanded portrait, `WizardExitConfirm`); letting either veto recovery would strand the user, since neither holds a lock to release and the portrait sets no `pointer-events: auto` of its own.

**Dropdown menus stop taking the lock** (`dropdown-menu-content.svelte`). `bits-ui` resolves `preventScroll ?? true`, so dropdowns locked by default despite needing it no more than its own sub-menus, popovers and selects, which all opt out. The story entry `⋯` menu is `sm:hidden` — mobile-only — and the most frequent trigger. `strategy` is pinned to `fixed` because it is otherwise derived from `preventScroll`, so opting out would silently change positioning inside the scrolling story container too.

The comment in `scrollLock.ts` claiming dropdown menus never lock was wrong for the pinned `bits-ui`, and was the stated justification for the selector — corrected, since the next person to edit that list would have reasoned from it.

## For the reviewer

**Verification is incomplete, deliberately so.** The suite is `environment: 'node'` with no DOM, so only the style predicate has unit tests (5 of them, weighted toward the half-locked states the `afterTick` gap produces).

| Claim | Status |
|---|---|
| `isBodyLocked` predicate | ✅ unit tested |
| `<html>` stays hit-testable under the lock | 🟡 Chromium desktop, real input events — **not** Android WebView |
| `wheel` reaches document under the lock | 🟡 same |
| Listener-driven recovery in the app | ❌ unverified |
| Closing / stacked overlay behavior | ❌ unverified |
| Dropdown positioning after `strategy="fixed"` | ❌ unverified |

The middle rows are engine-level evidence for a CSS premise, gathered on a synthetic page. They say the approach is viable; they say nothing about whether this app's wiring is right.

**The load-bearing risk**: if `<html>` is *not* hit-testable under the lock in Android WebView, this approach fails entirely rather than partially. Worth settling on-device before this lands.

Behaviour change to expect: content behind an open dropdown menu now scrolls. Dialogs, alert dialogs and drawers keep locking, so no full-screen modal is affected.

Planning artifacts (proposal, spec, design, tasks, and two rounds of review with responses) live in the `aventuras-specs` store under `recover-orphaned-scroll-lock`; they are not part of this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery from stuck page scroll locks, including recovery after user interaction.
  - Updated overlay handling to more reliably identify and release abandoned scroll locks.
  - Dropdown menus and submenus no longer interfere with page scrolling by default and maintain stable positioning.
  - Improved handling of overlays while they are closing to prevent premature scroll-lock recovery.

- **Internal Improvements**
  - Improved scroll-lock detection and coordination across menus and overlays.
  - Added automated coverage for scroll-lock state detection and closing-overlay behavior.

### DIFF
--- a/src/lib/utils/scrollLock.ts
+++ b/src/lib/utils/scrollLock.ts
@@ -13,6 +13,8 @@
  * Nothing here is a substitute for closing a modal properly. It is the net under it.
  */
 
+import { MODAL_CLOSE_TRANSITION_MS } from '$lib/constants/layout'
+
 /** Marks menu content that has opted out of the body lock, so it is never mistaken for an owner. */
 export const NO_SCROLL_LOCK_ATTR = 'data-no-scroll-lock'
 
@@ -20,22 +22,52 @@ export const NO_SCROLL_LOCK_ATTR = 'data-no-scroll-lock'
  * Everything in this stack that legitimately holds a body lock, as it appears in the DOM.
  *
  * `bits-ui` resolves `preventScroll ?? true`, so dialog, alert-dialog, context-menu and
- * dropdown/menu lock; popover, select, tooltip and sub-menus opt out. `vaul` locks its own way.
+ * dropdown/menu lock. Popover, select and tooltip opt out upstream; a sub-menu shares
+ * `MenuContentState` and so looks identical here, and is excluded by the same marker the
+ * app's own dropdowns carry. `vaul` locks its own way.
  *
- * Matched on presence, not `[data-state="open"]`: an owner is mounted with its element and
- * still holds the lock while animating closed. `[data-state]` itself separates a library-owned
- * overlay from a hand-rolled one with the same role, which holds no lock and must not veto.
+ * `[data-state]` is what separates a library-owned overlay from a hand-rolled one with the
+ * same ARIA role, which holds no lock and must not be able to veto recovery.
  */
-const OPEN_OVERLAY_SELECTOR = [
-  '[role="dialog"][data-state]',
-  '[role="alertdialog"][data-state]',
-  `[role="menu"][data-state]:not([${NO_SCROLL_LOCK_ATTR}])`,
-  '[data-vaul-drawer][data-state]',
-].join(', ')
+const OVERLAY_CLAUSES = [
+  '[role="dialog"]',
+  '[role="alertdialog"]',
+  `[role="menu"]:not([${NO_SCROLL_LOCK_ATTR}])`,
+  '[data-vaul-drawer]',
+]
 
-/** Is anything on screen entitled to be holding the body lock right now? */
+const OPEN_OVERLAY_SELECTOR = OVERLAY_CLAUSES.map((c) => `${c}[data-state="open"]`).join(', ')
+const CLOSING_OVERLAY_SELECTOR = OVERLAY_CLAUSES.map((c) => `${c}[data-state="closed"]`).join(', ')
+
+/** When each closing overlay was first seen, so the grace below is measured from somewhere. */
+const closingSince = new WeakMap<Element, number>()
+
+/** Is a closing overlay still within the window where it may hold its lock? */
+export function isWithinCloseGrace(since: number, now: number): boolean {
+  return now - since < MODAL_CLOSE_TRANSITION_MS
+}
+
+/**
+ * Is anything on screen entitled to be holding the body lock right now?
+ *
+ * A closing overlay still owns its lock — the owner unmounts with the element, not with the
+ * state flip — so it counts, but only until its exit has had time to finish. `bits-ui` drops
+ * the element from an animation callback, and a WebView that never delivers one would
+ * otherwise leave a node at `data-state="closed"` vetoing recovery for the rest of the session.
+ */
 function hasOpenOverlay(): boolean {
-  return document.querySelector(OPEN_OVERLAY_SELECTOR) !== null
+  if (document.querySelector(OPEN_OVERLAY_SELECTOR) !== null) return true
+
+  const now = Date.now()
+  for (const el of document.querySelectorAll(CLOSING_OVERLAY_SELECTOR)) {
+    const since = closingSince.get(el)
+    if (since === undefined) {
+      closingSince.set(el, now)
+      return true
+    }
+    if (isWithinCloseGrace(since, now)) return true
+  }
+  return false
 }
 
 /**
--- a/src/lib/utils/scrollLock.ts
+++ b/src/lib/utils/scrollLock.ts
@@ -13,18 +13,31 @@
  * Nothing here is a substitute for closing a modal properly. It is the net under it.
  */
 
+/** Marks menu content that has opted out of the body lock, so it is never mistaken for an owner. */
+export const NO_SCROLL_LOCK_ATTR = 'data-no-scroll-lock'
+
 /**
  * Everything in this stack that legitimately holds a body lock, as it appears in the DOM.
  *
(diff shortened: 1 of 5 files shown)


src/lib/utils/scrollLock.ts
```diff
@@ -13,18 +13,24 @@
...
 const OPEN_OVERLAY_SELECTOR = [
-  '[role="dialog"][data-state="open"]',
-  '[role="alertdialog"][data-state="open"]',
-  '[role="menu"][data-state="open"]',
-  '[data-vaul-drawer][data-state="open"]',
+  '[role="dialog"][data-state]',
+  '[role="alertdialog"][data-state]',
+  `[role="menu"][data-state]:not([${NO_SCROLL_LOCK_ATTR}])`,
+  '[data-vaul-drawer][data-state]',
 ].join(', ')
```
a-frazier: **Bare `[data-state]` can permanently disable orphan-lock recovery**

Switching from `[data-state="open"]` to bare `[data-state]` makes these selectors match an overlay node left mounted at `data-state="closed"`. On the Android WebView this PR targets, a bits-ui/vaul overlay whose exit `animationend`/`transitionend` never fires (a known WebView hazard) stays in the DOM at `data-state="closed"`. `hasOpenOverlay()` then returns `true` forever, so `releaseOrphanScrollLock()` early-returns on every trigger and the body stays permanently `pointer-events: none` — the exact freeze this net exists to break. The prior `[data-state="open"]` selector did not match a stalled-closed node, so it did not have this failure mode.

src/lib/utils/scrollLock.ts
```diff
@@ -13,18 +13,24 @@
...
+ * overlay from a hand-rolled one with the same role, which holds no lock and must not veto.
  */
 const OPEN_OVERLAY_SELECTOR = [
-  '[role="dialog"][data-state="open"]',
-  '[role="alertdialog"][data-state="open"]',
-  '[role="menu"][data-state="open"]',
-  '[data-vaul-drawer][data-state="open"]',
+  '[role="dialog"][data-state]',
+  '[role="alertdialog"][data-state]',
+  `[role="menu"][data-state]:not([${NO_SCROLL_LOCK_ATTR}])`,
```
a-frazier: **An open bits-ui sub-menu falsely vetoes recovery**

`data-no-scroll-lock` is added only in `dropdown-menu-content.svelte`, not `dropdown-menu-sub-content.svelte`. bits-ui `MenuSubContent` shares `MenuContentState`, so it renders `role="menu"` `data-state="open"` with no lock and no opt-out attribute — so it matches `[role="menu"][data-state]:not([data-no-scroll-lock])` and makes `hasOpenOverlay()` return `true` while holding no lock. While such a sub-menu is open, `releaseOrphanScrollLock()` refuses to clear a genuinely orphaned lock. The comment just above even asserts "sub-menus opt out," which contradicts this selector. Currently latent (no `Sub` call sites in `src/`), but `Sub`/`SubTrigger`/`SubContent` are exported from `dropdown-menu/index.ts` for use.

a-frazier (commented):
Code review found 2 correctness issues in `scrollLock.ts`, both around the switch to a bare `[data-state]` selector. Details inline.

src/lib/utils/scrollLock.ts
collegatore: Confirmed and fixed in feb59afe — presence unmount runs from `#afterAnimations.run(...)`, so a stalled exit really can leave a node at `data-state="closed"` vetoing forever. Rather than revert to `[data-state="open"]` (which reopens the release-under-a-still-exiting-modal case this replaced), entitlement for a closing overlay now expires after `MODAL_CLOSE_TRANSITION_MS`, measured from when it was first seen; `isWithinCloseGrace` is unit tested including the stalled case.

src/lib/utils/scrollLock.ts
collegatore: Confirmed and fixed in feb59afe — `menu-sub-content.svelte` builds on the same `MenuContentState`, so it renders `role="menu"` + `data-state` while passing `preventScroll={false}`. `dropdown-menu-sub-content.svelte` now emits the opt-out marker, and the misleading "sub-menus opt out" comment is corrected to say they look identical here and are excluded by the marker.

a-frazier (approved):

src/lib/utils/scrollLock.ts
```diff
@@ -13,27 +13,82 @@
...
+  if (document.querySelector(OPEN_OVERLAY_SELECTOR) !== null) return true
+
+  const now = Date.now()
+  for (const el of document.querySelectorAll(CLOSING_OVERLAY_SELECTOR)) {
+    const since = closingSince.get(el)
+    if (since === undefined) {
+      closingSince.set(el, now)
+      return true
+    }
+    if (isWithinCloseGrace(since, now)) return true
```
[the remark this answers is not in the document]
collegatore: Confirmed and fixed in 78822739 — a close/reopen/close on the same element reused the first close's timestamp, so the second close inherited an expired grace and released the lock while still exiting. Took the fix (clearing entries for open overlays first), and moved the bookkeeping into `hasEntitledOwner`, keyed on any object rather than `Element`, so the regression is covered by unit tests despite the DOM-less test env — including the close → reopen → close sequence you asked for.

PR #490, opened 2026-09-07
Variable-length retrieval query stack (1/3)

failerko:
## Summary

PR 1 of three implementing the [2026-09-06 query-stack design](docs/explorations/2026-09-06-retrieval-query-stack.md) (`bae02bc3`). Canon is [`retrieval.md → Query construction`](docs/memory/retrieval.md#query-construction--the-query-stack); this PR is code catching up to it.

- **Q3 becomes the piggyback summary.** `metadata.summary` splits out of Q2 into a query of its own with its own weight. Q2 is now purely structural — deterministic and free, so it never fails because a model had a bad turn. The heuristic prose extract that used to be Q3 is deleted.
- **The stack becomes variable-length** — three to six presence-derived slots. Slots 4-6 are Q4, classifier-emitted queries, capped at three, empty-filtered, deduped and length-capped before embedding. `QueryWeights` goes to four keys with `w_direct` **pooled** across every emitted Q4, so emission volume does not become influence.
- **The probe capture follows** at `CAPTURE_VERSION` 6: `queries` and per-row `sims` are lists, and `replay.ts` derives its slot order from the capture instead of a hardcoded triple, making simulator/production parity structural rather than conventional.
- **The fallback classifier is told to write the summary** its schema always allowed — it had no `.describe()` and no prompt instruction, so the field it now owes a fifth of the blend to was one the model had little reason to fill.

## Two things a reviewer should know

**Q4 has no producer yet — that is PR 2.** The mechanism, cap/dedupe/filter rules and pooled weight ship here exercised by tests only. That is intentional, not dead code.

**So until PR 2 lands, the live blend is action 0.40 / digest 0.33 / summary 0.27**, not the nominal 0.30/0.25/0.20/0.25 — `w_direct` renormalizes away every turn. Expect ranking to shift when PR 2 merges; that is the moment to calibrate, not a regression.

## The stack

| PR | Owns | Migration |
| --- | --- | --- |
| **1 (this)** | Variable-length stack, Q3 = summary, prose extract deleted, capture v6 | none |
| 2 | Q4 end-to-end on the piggyback path + redundancy + capture v7 + entry card | none |
| 3 | Fallback parity: context expansion, Q4 schema field, `classifierContextEntries` | `0013` settings backfill |

PR 2 must not source `emittedQueries` before this PR's capture reshape — `RetrievalParams.query` already carries the field, so a capture written at v5 would silently drop every Q4 sim and diverge from replay with no error.

## Test Plan

- [x] `pnpm test:run` — 444 files / 4671 tests pass (unit + storybook)
- [x] `pnpm typecheck`, `pnpm lint`, `pnpm lint:docs` clean
- [x] Every task through spec-compliance then code-quality review, with independent mutation batteries
- [ ] Manual smoke: `pnpm db:seed && pnpm desktop`, open a story, submit a turn, confirm a reply and no retrieval error

## Summary by CodeRabbit

- **New Features**
  - Retrieval supports up to six query signals, including generated follow-up queries.
  - Recent turn summaries improve context-aware results and identify relevant people, places, and things by name.

- **Improvements**
  - Ranking blends action, structural, summary, and direct-match signals more consistently.
  - Capture and replay support variable query sets with aligned similarity results.

- **Bug Fixes**
  - Invalid or outdated captured data is now rejected with clearer validation errors.


lib/retrieval/queries.ts
```diff
@@ -42,45 +53,52 @@ function structuralDigest(input: QueryStackInput): string {
...
-  const embedTexts = [q1, q2, q3].filter((_, i) => presence[i]).map((q) => q.text)
+  // An absent fixed slot is recorded but not embedded: the probe renders it, and
+  // the blend re-normalizes over the present ones rather than spending a share on noise.
+  const specs: QuerySpec[] = [q1, q2, q3, ...emittedSpecs(input.emittedQueries ?? [])]
+  const slots = specs.map((q) => QUERY_SLOT_OF_SOURCE[q.source])
+  const presence = specs.map((q) => nonEmpty(q.text))
+  const embedTexts = specs.filter((_, i) => presence[i]).map((q) => q.text)
 
-  return { q1, q2, q3, presence, embedTexts }
+  return { specs, slots, presence, embedTexts, q1, q2, q3 }
```
a-frazier: **Maintainability (not a correctness bug):** `QueryStack` keeps `q1`/`q2`/`q3` alongside `specs`, but after this PR production reads only `specs`/`slots`/`presence`/`embedTexts` — `payload.ts`, `run.ts` and `ranker.ts` all consume those. `q1`/`q2`/`q3` are now referenced only by tests, so the first three slots are represented two ways on the same object.

A future change to how a fixed slot is built has to be mirrored in both the `specs` array and the `q1`/`q2`/`q3` assignments or they silently disagree. Dropping `q1`/`q2`/`q3` from the return and having the tests index `specs[0..2]` removes the duplication.

_(Reported by an automated review pass; posting inline for visibility.)_

a-frazier (approved):

PR #491, opened 2026-09-07
Prompt packs: export and import as a directory of Markdown files

collegatore:
## Why

The built-in editor is the only way to read or change a pack's ~80 templates: one at a
time, with weak search and no diff. A set of edits cannot be reviewed, carried across a
release, or kept anywhere but the database. When a new version ships better defaults, a
user who has customised them has no way to combine the two other than re-typing.

Exporting a pack as a directory hands that job to tools that already do it well — an
editor, `ripgrep`, and git — and turns the release-to-release merge into an ordinary
vendor-branch workflow.

## What changed

- **Export a pack as a directory.** One `.md` file per stored row, in a folder named for
  its display group. `pack.yaml` at the root carries the format version, app version, pack
  metadata and custom variables.
- **Export the shipped baseline**, a separate action in the built-in pack's Pack Settings.
  It reads `PROMPT_TEMPLATES` rather than any pack's rows, so it is unedited by
  construction and safe as a merge base.
- **Import a directory back**, as a new pack or over an existing one. Replace semantics,
  matching `.prompt.json`. The built-in pack stays excluded as a target.
- **Generated `_reference/`** the importer ignores: the system and runtime variable
  registries, a template-to-variable index and its reverse, and a per-template status
  report naming what carries an edit.
- **Generated `ABOUT.md` and `.gitattributes`**, explaining the format and pinning LF.
- Pack cards now group all four transfer actions under one import/export dropdown, matching
  the story header's menu, instead of a row of near-identical icons.

Desktop only — `tauri-plugin-dialog`'s folder picker is `FolderPickerNotImplemented` on
mobile. The `.prompt.json` path is untouched and remains the Android route.

## Notes for review

**The filename is the identity; the folder is presentation.** Groups come from
`TEMPLATE_GROUP_MAP`, which is code and changes between releases. Were the path the
identity, a regrouped template would read as a delete plus an add of a file the user may
also have edited. As a stem it is a rename, which git detects.

**Determinism is load-bearing.** Two exports of unchanged content must be byte-identical or
git cannot merge them — hence LF endings, one trailing newline, fixed key order, variables
sorted, and deliberately no timestamp.

**Imports write rows as their own baseline** (`isBaseline: true`), as `updatePackFromFile`
does. The file *is* the pack's baseline; the user's edits live in git. With `false`, every
row reads as edited and `summarizeUpdate` warns about discarding ~80 edits on every import
— a warning that always fires stops being read.

**Liquid failures are asymmetric, and this is documented rather than fixed.** Structural
faults (unclosed tag, mismatched end tag, unknown filter) are refused whole, each file named
with line and column, nothing written. An unknown *variable* name is not a fault: Liquid
renders it empty and nothing reports it, ever. `ABOUT.md` leads with that, since it is what
a bad merge produces and no check can catch.

**`yaml` moved from devDependencies to dependencies.** `pack.yaml` is hand-edited during a
merge, so comments and unquoted multi-line strings were worth the dependency over JSON.

One capability change: `fs:allow-read-dir`.

## Verification

`npm run check` 0 errors · `npm run lint` 0 errors · `npm test` 1437 passed, 9 skipped
(79 new tests across 7 files).

A full desktop cycle was walked by hand: export the shipped baseline, commit it, edit a
template outside the app, import as a new pack, confirm no modified badge and that a bound
story generates from the edited text.

Three findings from an automated review are fixed in this branch, the first of which was a
genuine data-loss path: on a case-insensitive filesystem the prune planner could schedule a
file it had just written under different casing, so a *successful* export silently lost a
template.

## Summary by CodeRabbit

* **New Features**
  * Added folder-based import and export for prompt packs.
  * Added options to update packs from folders and export shipped prompts to a folder.
  * Added confirmation before exporting into folders containing existing files, including files that may be removed.
  * Separated file and folder transfer options.
  * Folder-transfer options are hidden on unsupported platforms; file-based transfer remains available.

* **Bug Fixes**
  * Improved import validation messages with file paths and clearer replacement guidance.
  * Prevented template files differing only by capitalization from being imported as duplicates.
  * Improved handling of concurrent folder transfers and canceled export confirmations.

### DIFF
--- a/src/lib/components/vault/prompts/PromptPackCard.svelte
+++ b/src/lib/components/vault/prompts/PromptPackCard.svelte
@@ -41,15 +41,24 @@
   const canReplace = $derived(!pack.isDefault && !!onUpdateFromFile)
 </script>
 
-<button type="button" class="w-full text-left" {onclick}>
+<div class="relative">
   <Card
     class="group hover:border-primary/50 h-full cursor-pointer transition-colors {pack.isDefault
       ? 'border-dashed'
       : ''}"
   >
     <CardContent class="flex h-full flex-col p-4">
+      <!-- Covers the card so the whole surface opens the pack, while the action menu stays a
+           sibling: a menu trigger nested in a <button> breaks focus and activation. -->
+      <button
+        type="button"
+        class="absolute inset-0 z-0 h-full w-full cursor-pointer rounded-xl text-left"
+        aria-label="Open {pack.name}"
+        {onclick}
+      ></button>
+
       <div class="flex items-start justify-between">
-        <div class="min-w-0 flex-1">
+        <div class="pointer-events-none relative z-10 min-w-0 flex-1">
           <div class="flex flex-wrap items-center gap-2">
             <h3 class="truncate font-semibold">{pack.name}</h3>
             {#if usageCount > 0}
@@ -60,7 +69,7 @@
             {#if pack.description}{stripToPlainText(pack.description)}{:else}&nbsp;{/if}
           </p>
         </div>
-        <div class="flex shrink-0 items-center gap-0.5">
+        <div class="relative z-10 flex shrink-0 items-center gap-0.5">
           <DropdownMenu.Root bind:open={transferMenuOpen}>
             <DropdownMenu.Trigger>
               {#snippet child({ props })}
@@ -138,4 +147,4 @@
       </div>
     </CardContent>
   </Card>
-</button>
+</div>
--- a/src/lib/components/vault/prompts/PromptPackCard.svelte
+++ b/src/lib/components/vault/prompts/PromptPackCard.svelte
@@ -3,8 +3,10 @@
   import { Card, CardContent } from '$lib/components/ui/card'
   import { Badge } from '$lib/components/ui/badge'
   import { Button } from '$lib/components/ui/button'
-  import { Upload, Trash2, Lock, RefreshCw } from '@lucide/svelte'
+  import * as DropdownMenu from '$lib/components/ui/dropdown-menu'
+  import { Trash2, Lock, ArrowUpDown, ChevronDown, FileJson, FolderOpen } from '@lucide/svelte'
   import { stripToPlainText } from '$lib/utils/markdown'
+  import { supportsDirectoryTransfer } from '$lib/services/packs/directory/support'
 
   interface Props {
     pack: PresetPack
@@ -12,12 +14,31 @@
     usageCount: number
     onclick: () => void
     onExport?: () => void
+    onExportDirectory?: () => void
     onUpdateFromFile?: () => void
+    onUpdateFromDirectory?: () => void
     onDelete?: () => void
   }
 
-  let { pack, modifiedCount, usageCount, onclick, onExport, onUpdateFromFile, onDelete }: Props =
-    $props()
+  let {
+    pack,
+    modifiedCount,
+    usageCount,
+    onclick,
+    onExport,
+    onExportDirectory,
+    onUpdateFromFile,
+    onUpdateFromDirectory,
+    onDelete,
+  }: Props = $props()
+
+  const canUseDirectories = supportsDirectoryTransfer()
+
+  let transferMenuOpen = $state(false)
+
+  // The built-in pack is rewritten from the app's own templates on every launch, so it is
+  // never a replacement target.
+  const canReplace = $derived(!pack.isDefault && !!onUpdateFromFile)
 </script>
 
 <button type="button" class="w-full text-left" {onclick}>
@@ -40,34 +61,56 @@
           </p>
         </div>
         <div class="flex shrink-0 items-center gap-0.5">
-          {#if onExport}
-            <Button
-              variant="ghost"
-              size="icon"
-              class="h-8 w-8 transition-opacity sm:opacity-0 sm:group-hover:opacity-100"
-              onclick={(e: MouseEvent) => {
-                e.stopPropagation()
-                onExport?.()
-              }}
-              title="Export pack"
-            >
-              <Upload class="h-4 w-4" />
-            </Button>
-          {/if}
-          {#if !pack.isDefault && onUpdateFromFile}
-            <Button
-              variant="ghost"
-              size="icon"
-              class="h-8 w-8 transition-opacity sm:opacity-0 sm:group-hover:opacity-100"
-              onclick={(e: MouseEvent) => {
-                e.stopPropagation()
-                onUpdateFromFile?.()
-              }}
-              title="Update from file"
-            >
-              <RefreshCw class="h-4 w-4" />
-            </Button>
-          {/if}
+          <DropdownMenu.Root bind:open={transferMenuOpen}>
+            <DropdownMenu.Trigger>
+              {#snippet child({ props })}
+                <Button
+                  {...props}
+                  variant="ghost"
+                  size="icon"
+                  class="h-8 w-auto gap-0.5 px-1.5 transition-opacity sm:opacity-0 sm:group-hover:opacity-100 {transferMenuOpen
+                    ? 'sm:opacity-100'
+                    : ''}"
+                  onclick={(e: MouseEvent) => {
+                    e.stopPropagation()
+                    transferMenuOpen = !transferMenuOpen
+                  }}
+                  title="Import / Export pack"
+                >
+                  <ArrowUpDown class="h-4 w-4" />
+                  <ChevronDown class="h-3 w-3" />
+                </Button>
+              {/snippet}
+            </DropdownMenu.Trigger>
+            <DropdownMenu.Content align="end" onclick={(e: MouseEvent) => e.stopPropagation()}>
+              <DropdownMenu.Label>Export</DropdownMenu.Label>
+              <DropdownMenu.Item onclick={() => onExport?.()}>
+                <FileJson class="text-accent-400 h-4 w-4" />
+                Pack file (.prompt.json)
+              </DropdownMenu.Item>
+              {#if canUseDirectories}
+                <DropdownMenu.Item onclick={() => onExportDirectory?.()}>
+                  <FolderOpen class="h-4 w-4 text-blue-400" />
+                  Folder of Markdown
+                </DropdownMenu.Item>
+              {/if}
+
+              {#if canReplace}
+                <DropdownMenu.Separator />
+                <DropdownMenu.Label>Replace this pack from</DropdownMenu.Label>
+                <DropdownMenu.Item onclick={() => onUpdateFromFile?.()}>
+                  <FileJson class="text-accent-400 h-4 w-4" />
+                  Pack file (.prompt.json)
+                </DropdownMenu.Item>
+                {#if canUseDirectories && onUpdateFromDirectory}
+                  <DropdownMenu.Item onclick={() => onUpdateFromDirectory?.()}>
+                    <FolderOpen class="h-4 w-4 text-blue-400" />
+                    Folder of Markdown
+                  </DropdownMenu.Item>
+                {/if}
+              {/if}
+            </DropdownMenu.Content>
+          </DropdownMenu.Root>
           {#if !pack.isDefault && onDelete}
             <Button
               variant="ghost"
--- a/src/lib/components/vault/prompts/PromptPackCard.svelte
+++ b/src/lib/components/vault/prompts/PromptPackCard.svelte
@@ -70,6 +70,8 @@
           </p>
         </div>
         <div class="relative z-10 flex shrink-0 items-center gap-0.5">
+          <!-- No onclick on the trigger: its own handlers arrive through the spread props, and
+               a second toggle on click opens then immediately closes it on touch. -->
           <DropdownMenu.Root bind:open={transferMenuOpen}>
             <DropdownMenu.Trigger>
               {#snippet child({ props })}
@@ -80,10 +82,6 @@
                   class="h-8 w-auto gap-0.5 px-1.5 transition-opacity sm:opacity-0 sm:group-hover:opacity-100 {transferMenuOpen
                     ? 'sm:opacity-100'
                     : ''}"
-                  onclick={(e: MouseEvent) => {
-                    e.stopPropagation()
-                    transferMenuOpen = !transferMenuOpen
-                  }}
                   title="Import / Export pack"
                 >
                   <ArrowUpDown class="h-4 w-4" />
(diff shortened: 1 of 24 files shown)


src/lib/components/vault/prompts/PromptPackCard.svelte
```diff
@@ -40,34 +61,56 @@
...
-              }}
-              title="Update from file"
-            >
-              <RefreshCw class="h-4 w-4" />
-            </Button>
-          {/if}
+          <DropdownMenu.Root bind:open={transferMenuOpen}>
+            <DropdownMenu.Trigger>
+              {#snippet child({ props })}
+                <Button
```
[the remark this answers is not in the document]
collegatore: Moved the action menu out of the card button: the card is now a positioned container with a stretched button behind the content, so the menu trigger is its sibling rather than a nested <button>.

src/lib/services/packs/directory/layout.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Whether this id survives being written as a filename and read back.
+ *
+ * `PackTemplateSchema` accepts any non-empty string, so a hand-written `.prompt.json` can
+ * carry an id holding a path separator. Written out it would gain a folder and come back as
+ * only its last segment — a different template, silently. Rather than encode ids and make
+ * every filename harder to read, an export refuses the ones that would not survive.
+ */
+export function isRoundTrippableId(templateId: string): boolean {
+  return templateIdFromPath(templatePath(templateId)) === templateId
```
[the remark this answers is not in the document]
collegatore: Renamed isRoundTrippableId to isWritableTemplateId and extended it to Win32-forbidden characters, control characters, reserved device names (CON, COM1, ...) and a trailing dot or space, so buildTree refuses them before writeTree can fail.

src/lib/services/packs/directory/tree.ts
```diff
@@ -0,0 +1,150 @@
...
+    )
+  }
+
+  const tree: Tree = new Map()
+
+  const contentById = new Map<string, string>()
+  for (const template of input.templates) {
+    tree.set(templatePath(template.templateId), serializeTemplate(template.content))
+    contentById.set(template.templateId, template.content)
+  }
```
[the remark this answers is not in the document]
collegatore: buildTree now refuses two template ids that differ only in capitalisation, through the same unwritable mechanism - they are one file on Windows and macOS, so one would silently overwrite the other.

src/lib/services/packs/import-export.ts
```diff
@@ -69,6 +101,93 @@ class ImportExportService {
...
+  /** Where an existing tree's files sit, and which of them this export would leave behind. */
+  private async planWrite(root: string, tree: Tree): Promise<DirectoryExportPlan> {
+    let existingPaths: string[] = []
+    try {
+      existingPaths = (await readTree(root)).allPaths
+    } catch (e) {
+      console.error('[ImportExportService] Failed to read the chosen folder:', e)
+    }
+
+    const isPreviousExport = existingPaths.some((p) => p.replace(/\\/g, '/') === PACK_FILE)
```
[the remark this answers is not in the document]
collegatore: planWrite now requires the root pack.yaml to parse as a pack file this version can read before enabling pruning; a folder that cannot be read prunes nothing and asks for confirmation first.

src/lib/services/packs/directory/tree.ts
```diff
@@ -0,0 +1,165 @@
...
+  // paths, not as a template silently missing from a successful export.
+  const written = new Set([...tree.keys()].map((path) => path.toLowerCase()))
+  const normalized = (path: string) => path.replace(/\\/g, '/')
+
+  return existingPaths
+    .map(normalized)
+    .filter((path) => !written.has(path.toLowerCase()))
+    .filter(
+      (path) =>
+        classifyPath(path) === 'template' ||
```
a-frazier, line 161: **Data loss: a plain re-export silently deletes user-authored `.md` files kept inside a group folder.**

`classifyPath` returns `'template'` for *any* `.md` file below the root whose path has no `_`-prefixed segment, so this filter sweeps in files the user put there themselves (e.g. `Story Generation/ideas.md` notes committed next to the prompts). On a re-export into a folder that is a recognized prior export, `planWrite` in `import-export.ts` sets `isPreviousExport = true` (a parseable `pack.yaml`), which makes `prunePaths = planPrune(...)` and `needsConfirmation = false` — so those files are passed to `writeTree`'s `remove()` loop with no dialog.

`ABOUT.md` only promises that *root-level* `.md` files and `_reference/` are safe; a notes file beside the prompts is neither protected nor mentioned, which actively misleads the user about what survives.

Consider restricting the prune to paths the export has previously written (e.g. persist the last-written manifest, or only prune ids that resolve to a currently-shipped/known template), or at minimum surface a confirmation listing what will be removed even when the folder is a known prior export.

src/lib/services/packs/directory/io.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/** Every file below `root`, as paths relative to it. */
+export async function listFiles(root: string, relativeDir = ''): Promise<string[]> {
+  const entries = await readDir(relativeDir ? joinPath(root, relativeDir) : root)
+  const files: string[] = []
+
+  for (const entry of entries) {
+    const relative = relativeDir ? `${relativeDir}/${entry.name}` : entry.name
+    if (entry.isDirectory) files.push(...(await listFiles(root, relative)))
+    else if (entry.isFile) files.push(relative)
```
a-frazier: **`listFiles` recurses unconditionally into every subdirectory, including `.git` and `node_modules`.**

`readTree` (import + `planWrite`) calls `listFiles(root)`, which descends into every directory. The workflow this feature is built for keeps the exported tree in a git repo ("an ordinary vendor-branch workflow"; *"If you use git, keep that file"* in `ABOUT.md`), so on every folder import and every export into that repo this walks the entire `.git` tree — one async `readDir` IPC round-trip per directory, thousands of loose-object paths accumulated into `allPaths`. Expect multi-second stalls and a large memory spike on a normal-sized repo.

`classifyPath` already discards anything under a dot/underscore-prefixed directory, so descending into them is pure waste. Skipping `entry.name.startsWith('.')` (and ideally `node_modules`) during traversal would fix it.

src/lib/components/vault/prompts/PromptPackList.svelte
```diff
@@ -236,6 +301,35 @@
...
+      <Button
+        variant="outline"
+        onclick={() => {
+          pendingExport = null
+        }}>Cancel</Button
+      >
+      <Button onclick={confirmExportIntoUsedFolder}>Export here</Button>
+    </ResponsiveModal.Footer>
+  </ResponsiveModal.Content>
+</ResponsiveModal.Root>
```
a-frazier: **Duplicated with `PromptPackEditor.svelte`.** This "Export into this folder?" `ResponsiveModal` (comment + markup), the `confirmExportIntoUsedFolder` handler above, and the `try/catch` + toast in the export handler are near-verbatim copies of the same three blocks in `PromptPackEditor.svelte`. The overwrite warning already enumerates `ABOUT.md` / `.gitattributes`; the next wording or behavior change to the confirm flow has to be made in both places, and it's easy to update one and miss the other.

A shared `<ExportIntoUsedFolderDialog>` component (taking the pending plan + confirm/cancel callbacks) invoked from both call sites would remove the duplication.

src/lib/services/packs/directory/layout.ts
```diff
@@ -0,0 +1,152 @@
...
+    const existing = byId.get(templateId)
+    if (existing) existing.push(path)
+    else byId.set(templateId, [path])
+  }
+
+  const duplicates: DuplicateTemplate[] = []
+  const templates = new Map<string, string>()
+
+  for (const [templateId, found] of byId) {
+    if (found.length > 1) duplicates.push({ templateId, paths: [...found].sort() })
```
a-frazier, line 149: **Import has no case-insensitive stem-collision check, but export (`buildTree`) does.**

`buildTree` throws on two ids that "differ only in capitalisation" because they'd be one file on Windows/macOS. `collectTemplateFiles` keys `byId` by exact case, so on a case-sensitive filesystem a merged/hand-assembled tree with `Analysis/adventure.md` and `Memory/Adventure.md` imports cleanly as two rows, `adventure` and `Adventure`.

Exporting that pack later on Windows or macOS then makes `buildTree` throw `differ only in capitalisation`, and there's no UI to rename a stored row — the pack's directory export is permanently broken. Worth folding a lowercased-key collision check in here so the two directions agree.

src/lib/components/vault/prompts/PromptPackList.svelte
```diff
@@ -94,16 +103,67 @@
...
+
+      await importExportService.applyDirectoryExport(plan)
+      ui.showToast('Pack exported to folder', 'info')
+    } catch (e) {
+      console.error('Folder export failed:', e)
+      ui.showToast(`Export failed: ${errMessage(e)}`, 'error')
+    }
+  }
+
+  async function confirmExportIntoUsedFolder() {
```
a-frazier, line 135: **No in-flight guard on `handleExportPackDirectory` / `handleUpdateFromDirectory`** (contrast `PromptPackEditor.svelte`, which guards the equivalent action with `disabled={exportingBaseline}`).

A double-click on the "Folder of Markdown" menu item — or a slow disk — invokes `handleExportPackDirectory` twice: two folder pickers open, and if the user picks in both, two exports run. `writeTree` passes can interleave, and the second can prune files the first just wrote. Same exposure for `handleUpdateFromDirectory` (line 137). Add a `busy`/`exporting` `$state` flag and gate both handlers + the card menu items on it.

a-frazier (commented):
Automated code review (level: high). Five findings, inline.

- **Data loss (blocking):** `planPrune` + `planWrite` silently delete user-authored `.md` files kept inside a group folder on a plain re-export into a recognized prior export — no confirmation. `ABOUT.md` only promises root-level files are safe.
- **Performance:** `listFiles` recurses into `.git` / `node_modules` on every import and export-plan — multi-second stalls for the git-repo workflow the feature targets.
- **Correctness:** import lacks the case-insensitive stem-collision check that `buildTree` enforces on export, so a pack can import in a state that can never be re-exported.
- **Duplication:** the "Export into this folder?" modal + handler are copied near-verbatim between `PromptPackList.svelte` and `PromptPackEditor.svelte`.
- **Robustness:** list-view export/update actions have no in-flight guard, unlike the editor's.

src/lib/services/packs/directory/tree.ts
collegatore: Confirmed and fixed: anything to be deleted now forces the confirmation and is listed in it (needsConfirmation is true whenever prunePaths is non-empty), so a re-export into a recognised tree can no longer remove a nested file silently. ABOUT.md now says a group folder is no place for your own notes, instead of implying they are safe because root-level ones are.

src/lib/services/packs/directory/io.ts
collegatore: listFiles now skips dot-directories and node_modules during traversal, so .git is no longer walked on every import and export plan.

src/lib/services/packs/directory/layout.ts
collegatore: collectTemplateFiles now keys by lowercased stem, so two files differing only in case are refused on import exactly as buildTree refuses them on export — no more packs that import cleanly but can never be re-exported.

src/lib/components/vault/prompts/PromptPackList.svelte
collegatore: Extracted ExportIntoFolderDialog.svelte, used by both PromptPackList and PromptPackEditor; the duplicated modal markup and its wording now live in one place.

src/lib/components/vault/prompts/PromptPackList.svelte
collegatore: Added a directoryBusy flag guarding both handleExportPackDirectory and handleUpdateFromDirectory, so a second click cannot open a second picker while one is in flight.

src/lib/components/vault/prompts/PromptPackList.svelte
```diff
@@ -94,16 +108,79 @@
...
+      if (!candidate.validation.valid || !candidate.validation.pack) {
+        updateErrorSource = 'folder'
+        updateErrors = candidate.validation
+        return
+      }
+
+      await openUpdateConfirmation(pack, candidate.validation)
+    } finally {
+      directoryBusy = false
+    }
```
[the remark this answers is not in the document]
collegatore: Added a catch to handleUpdateFromDirectory (and to handleImportPackDirectory in VaultPanel, which the summary flagged and had no try at all) so a picker rejection toasts instead of going unhandled. Kept it in the handlers rather than folding it into pickAndValidateDirectory: a picker that will not open is not a folder that fails validation, and reporting it as "the selected folder contains errors" would be wrong.

src/lib/components/vault/prompts/PromptPackList.svelte
```diff
@@ -94,16 +108,83 @@
...
+    if (!pendingExport) return
+    const plan = pendingExport
+    pendingExport = null
+    try {
+      await importExportService.applyDirectoryExport(plan)
+      ui.showToast('Pack exported to folder', 'info')
+    } catch (e) {
+      console.error('Folder export failed:', e)
+      ui.showToast(`Export failed: ${errMessage(e)}`, 'error')
+    }
```
a-frazier, line 150: `directoryBusy` doesn't cover the actual write, so concurrent directory exports/writes are still possible.

`handleExportPackDirectory` sets `directoryBusy = true`, but when the plan needs confirmation it does `pendingExport = plan; return`, and the `finally` clears `directoryBusy = false` immediately — before the user has even seen the dialog. That's fine for the confirmation phase (the dialog is modal, so a second pack's button can't be clicked while it's open).

The gap is here in `confirmExportIntoUsedFolder`, which runs the actual disk write (`applyDirectoryExport` → `writeTree`, i.e. `mkdir`/`writeTextFile`/`remove`) and never touches `directoryBusy`. Once the user clicks "Export here," the modal closes and the UI is fully interactive again while the write is still in flight. A second `handleExportPackDirectory`/`handleUpdateFromDirectory` targeting the same folder can now interleave with this one — e.g. the second export's `readTree`/`planPrune` runs against a half-written tree, or the first export deletes a file the second just wrote, silently corrupting the exported directory.

For contrast, `PromptPackEditor.svelte`'s shipped-baseline export flow re-arms its busy flag inside its own `confirmExportIntoUsedFolder` (`exportingBaseline = true` at the top). This one doesn't replicate that, so the in-flight guard added in response to the earlier review comment doesn't actually cover the highest-risk step — the file deletion.

collegatore: Right — the guard covered planning but not the write. It is now held from the pick until the confirmation is answered (the finally only releases it when no dialog was opened), and released in confirmExportIntoUsedFolder's own finally after applyDirectoryExport; cancelling releases it too, via an explicit cancel handler instead of clearing the pending plan inline. Applied the same shape to the editor's shipped-baseline flow, which re-armed around its write and so was not exposed — having the two differ is what let this hide.

PR #492, opened 2026-09-07
Narrator reinforcement: make the turn message a pack template

collegatore:
## Why

Every narrator turn sends two prompts. The system prompt is a pack template a user can edit; the user message was not — `NarrativeService` built it in TypeScript and prefixed it to every turn, restating the player-agency rules the pack already carries, including `NEVER write my dialogue, decisions, or inner thoughts`.

A user who reworded that rule in their own pack did not change it. They got both versions, and the one they did not write sat closest to the generation point — an open contradiction the model then spent its reasoning arbitrating.

Whether repeating those rules helps is model-dependent: a capable model needs them once, a smaller local one may do better with them twice. That was a per-story judgement frozen into the application.

## What changes

- **A `narratorReinforcement` story setting** — `full`, `minimal`, `none` — in the Writing Style controls the wizard and Story Settings share. Defaults to `full`.
- **The narrator turn message becomes a pack template.** `adventure` and `creative-writing` gain the `userContent` half every other prompt already has. The templates branch on the level in Liquid, the way they already branch on `pov` and `tense`, so **the pack decides what each level says** — the app only chooses between them.
- **`NarrativeService` stops building it.** Takes both halves of `ContextBuilder.render()` instead of discarding the user half; `buildPrimingMessage`, `buildAdventurePriming` and `buildCreativeWritingPriming` are gone.
- **The reinforcement is reworded while it moves.** The old text branched on `pov == 'third'` and treated everything else as second person, so a **first-person adventure was handed second-person instructions**. All three points of view now have their own text in both modes, and the tense comes from the story.
- **A setting that cannot take effect says so.** Story Settings disables the control when neither prompt references the variable — checking the user half *regardless* of a custom system prompt, since a custom system prompt replaces the system half only.
- **`docs/architecture/prompts.md` corrected.** It stated the two narrative templates deliberately had no user half, and never mentioned the hardcoded message standing in for it. That sentence is why this was hard to find.

## Compatibility

Existing stories default to `full`, so none loses its reinforcement — but **the wording changes**, deliberately. Only first-person adventures change materially, and they change from wrong to right.

No migration: `StorySettings` is a JSON blob and an absent key reads as the default. Existing packs pick up the two new templates through `backfillMissingTemplates`, which only ever inserts, so edited templates are untouched.

## Also fixed

`createStory` omitted `targetLength` from the `writingStyle` object it builds, so a Response Length chosen in the wizard never reached the created story even though `ScenarioService.prepareStoryData` maps it. Found while adding the new key to the same literal.

## Testing

`npm run check` 0 errors · `npm test` 1390 passed · `npm run lint` 0 errors.

Template tests assert the properties a reworded template must still hold — every POV names its own person, no unsubstituted variable survives, the agency rule appears at `full` and not below, `none` renders empty — rather than pinning wording that is meant to be edited. The availability precedence has its own tests taking prompt bodies as plain arguments, so nothing is mocked.

Manually verified in the app: the control, wizard persistence, the Vault insert menu, and the outgoing message at each level.

## Known gap, deferred

`backfillMissingTemplates` runs only from `PackService.initialize`, which is guarded by `this.initialized` and therefore runs once at startup. Neither `applyImport` nor `updatePackFromFile` calls it, so a pack imported from a file has no row for a template added since that file was exported, until the next launch. It still renders correctly — `ContextBuilder.resolveTemplate` falls back to `default-pack` and then to the compiled-in baseline — but the Vault editor lists rows, so the template is invisible and uneditable in that session.

This change makes it easier to hit by adding two templates. Closing it means making the backfill callable outside `initialize` (ideally per-pack, since an import knows which pack it wrote) and running it after both import paths; the invariant to preserve is that backfill only ever inserts, which is what makes running it more often safe. Left out of this PR because it belongs to the import path, not to narrator prompts.

## Summary by CodeRabbit

* **New Features**
  * Added Narrator Reinforcement settings with Full, Minimal, and None options.
  * Made the setting available during story setup and preserved it in story configuration.
  * Applied reinforcement to narrative prompts according to the selected level.
  * Added reinforcement details to prompt previews and template testing.
  * Indicates when reinforcement is unavailable for selected templates.
  * Improved first-person narrative guidance for adventure stories.
* **Tests**
  * Added coverage for reinforcement rendering, prompt behavior, template detection, and supported narrative perspectives.

### DIFF
--- a/src/lib/services/ai/generation/NarrativeService.ts
+++ b/src/lib/services/ai/generation/NarrativeService.ts
@@ -148,6 +148,17 @@ export interface NarrativeWorldState extends WorldStateContext {
 }
 
 /**
+ * Prefix the turn message with the narrator reinforcement the pack rendered.
+ *
+ * Emptiness is decided on the trimmed render, but the untrimmed value is what is sent:
+ * a level whose branches all miss can still leave newlines behind, and trimming what goes
+ * out would change the text `full` has always sent.
+ */
+export function joinReinforcement(reinforcement: string, userPrompt: string): string {
+  return reinforcement.trim() ? `${reinforcement}\n\n${userPrompt}` : userPrompt
+}
+
+/**
  * Build a block containing chapter summaries for injection into the system prompt.
  * Per design doc: summarized entries are excluded from direct context,
  * but their summaries provide narrative continuity.
@@ -297,7 +308,7 @@ export class NarrativeService {
     })
 
     // Build system prompt via ContextBuilder pipeline
-    const { systemPrompt, primingMessage } = await this.buildPrompts(
+    const { systemPrompt, reinforcement } = await this.buildPrompts(
       story,
       worldState,
       tieredContextBlock,
@@ -315,7 +326,7 @@ export class NarrativeService {
       // Stream using the main narrative profile
       const stream = streamNarrative({
         system: systemPrompt,
-        prompt: `${primingMessage}\n\n${userPrompt}`,
+        prompt: joinReinforcement(reinforcement, userPrompt),
         signal,
       })
 
@@ -357,7 +368,7 @@ export class NarrativeService {
     log('generate', { entriesCount: entries.length })
 
     // Build system prompt via ContextBuilder pipeline
-    const { systemPrompt, primingMessage } = await this.buildPrompts(
+    const { systemPrompt, reinforcement } = await this.buildPrompts(
       story,
       worldState,
       tieredContextBlock,
@@ -371,13 +382,13 @@ export class NarrativeService {
 
     return generateNarrative({
       system: systemPrompt,
-      prompt: `${primingMessage}\n\n${userPrompt}`,
+      prompt: joinReinforcement(reinforcement, userPrompt),
       signal,
     })
   }
 
   /**
-   * Build system and priming prompts through the ContextBuilder pipeline.
+   * Build both halves of the prompt through the ContextBuilder pipeline.
    *
    * Creates a ContextBuilder from the story, adds runtime variables
    * (tiered context, chapter summaries, style guidance), then renders
@@ -390,7 +401,7 @@ export class NarrativeService {
     styleReview?: StyleReviewResult | null,
     retrievedChapterContext?: string | null,
     timelineFillResult?: TimelineFillResult | null,
-  ): Promise<{ systemPrompt: string; primingMessage: string }> {
+  ): Promise<{ systemPrompt: string; reinforcement: string }> {
     const mode = story?.mode ?? 'adventure'
 
     // Create ContextBuilder -- forStory auto-populates mode, pov, tense, genre,
@@ -433,6 +444,9 @@ export class NarrativeService {
         targetLength,
         lengthInstruction: formatLengthInstruction(targetLength, mode),
       })
+    }
+    if (!ctx.getContext().narratorReinforcement) {
+      ctx.add({ narratorReinforcement: story?.settings?.narratorReinforcement || 'full' })
     }
 
     // Add runtime variables for template rendering
@@ -483,6 +497,11 @@ export class NarrativeService {
 
     // Render system prompt — use per-story override when set, otherwise fall back to pack template
     let systemPrompt: string
+    // A custom system prompt replaces the system half only; the turn message still comes
+    // from the pack, so the user half is rendered either way.
+    const templateId = mode === 'creative-writing' ? 'creative-writing' : 'adventure'
+    const { system, user } = await ctx.render(templateId)
+
     const customPrompt = story?.settings?.customSystemPrompt
     if (customPrompt) {
       const rendered = templateEngine.render(customPrompt, ctx.getContext())
@@ -493,28 +512,17 @@ export class NarrativeService {
       }
       systemPrompt = rendered
     } else {
-      const templateId = mode === 'creative-writing' ? 'creative-writing' : 'adventure'
-      const { system } = await ctx.render(templateId)
       systemPrompt = system
     }
-
-    // Build priming message based on mode/pov/tense
-    const context = ctx.getContext()
-    const primingMessage = this.buildPrimingMessage(
-      mode,
-      (context.pov as string) ?? 'second',
-      (context.tense as string) ?? 'present',
-      (context.protagonistName as string) ?? 'the protagonist',
-    )
 
     log('buildPrompts complete', {
       mode,
       usingCustomPrompt: !!customPrompt,
       systemPromptLength: systemPrompt.length,
-      primingMessageLength: primingMessage.length,
+      reinforcementLength: user.length,
     })
 
-    return { systemPrompt, primingMessage }
+    return { systemPrompt, reinforcement: user }
   }
 
   /**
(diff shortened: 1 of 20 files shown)


src/lib/services/prompts/templates/narrative.ts
```diff
@@ -143,6 +143,23 @@ End with a natural opening for action, not a direct question.{% endif %}
 {% endif %}{% if tieredContextBlock != blank %}
 {{ tieredContextBlock }}
 {% endif %}{% if styleGuidance != blank %}{{ styleGuidance }}{% endif %}`,
+  userContent: `{%- case narratorReinforcement %}
+{%- when 'minimal' %}You are the narrator of this interactive adventure. I am the player controlling protagonist named {{ protagonistName }}.
+{%- when 'full' %}You are the narrator of this interactive adventure. I am the player controlling protagonist named {{ protagonistName }}. Write in {{ tense }} tense, {{pov}} person.
+    {%- case pov %} 
+      {%- when 'first' %} {% assign actionExample = 'I push open the heavy door' %} 
+      {%- when 'second' %} {% assign actionExample = 'You push open the heavy door' %} 
+      {%- when 'third' %} {% assign actionExample =  protagonistName | append: ' pushes open the heavy door' %}
```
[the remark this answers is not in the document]
collegatore: Added first-person branches to `<style_instruction>` and `<response_instruction>` in the adventure system template, plus a test asserting the system half and the turn message name the same person for all three povs.

src/lib/services/prompts/templates/narratorReinforcement.ts
```diff
@@ -0,0 +1,47 @@
...
+export const NARRATOR_REINFORCEMENT_VAR = 'narratorReinforcement'
+
+/**
+ * Whether a template would honour the setting at all.
+ *
+ * A bare mention rather than a `{{ }}` match: the level is branched on, so it appears in
+ * `{% if %}`, `{% case %}` or `{% unless %}` as readily as in an output tag.
+ */
+export function templateUsesNarratorReinforcement(content: string | null | undefined): boolean {
+  return !!content && new RegExp(`\\b${NARRATOR_REINFORCEMENT_VAR}\\b`).test(content)
```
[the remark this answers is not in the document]
collegatore: `templateUsesNarratorReinforcement` now strips `{% comment %}` blocks and `{% # %}` inline comments (including whitespace-controlled forms) before matching, with tests for the commented cases and one asserting the shipped template's own comment block is not over-stripped.

src/lib/services/prompts/templates/narratorReinforcement.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Whether a template would honour the setting at all.
+ *
+ * A bare mention rather than a `{{ }}` match: the level is branched on, so it appears in
+ * `{% if %}`, `{% case %}` or `{% unless %}` as readily as in an output tag. Comments are
+ * removed first, so a branch someone commented out does not report the setting as working.
+ */
+export function templateUsesNarratorReinforcement(content: string | null | undefined): boolean {
+  if (!content) return false
+  const active = content.replace(LIQUID_COMMENT, '')
+  return new RegExp(`\\b${NARRATOR_REINFORCEMENT_VAR}\\b`).test(active)
```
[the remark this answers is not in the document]
collegatore: `templateUsesNarratorReinforcement` now collects Liquid tags and output expressions and matches within those, so prose naming the variable no longer counts while `{% if %}`, `{% case %}` and `{% unless %}` still do. Added literal-text tests, including prose alongside a real branch.

failerko (approved):

src/lib/services/ai/generation/NarrativeService.ts
```diff
@@ -483,6 +497,11 @@ export class NarrativeService {
 
     // Render system prompt — use per-story override when set, otherwise fall back to pack template
     let systemPrompt: string
+    // A custom system prompt replaces the system half only; the turn message still comes
+    // from the pack, so the user half is rendered either way.
+    const templateId = mode === 'creative-writing' ? 'creative-writing' : 'adventure'
+    const { system, user } = await ctx.render(templateId)
```
a-frazier: `buildPrompts` now unconditionally calls `ctx.render(templateId)` for the pack's system-half template even when the story has a `customSystemPrompt` — then the result is discarded because `systemPrompt = rendered` (from the custom prompt) overwrites it right after. Previously this render only ran in the no-custom-prompt branch.

Net effect: two extra `database.getPackTemplate()` lookups plus a full Liquid render of a (large) system template, on every narrator turn, for any story using Story Settings' custom system prompt — pure wasted I/O and render work.

src/lib/services/prompts/templates/narrative.ts
```diff
@@ -324,6 +358,39 @@ End at a natural narrative beat.{% endif %}
 {% endif %}{% if tieredContextBlock != blank %}
 {{ tieredContextBlock }}
 {% endif %}{% if styleGuidance != blank %}{{ styleGuidance }}{% endif %}`,
+  userContent: `{%- if narratorReinforcement == 'minimal' %}You are a skilled fiction writer. I am the author directing the story. Write what I ask for.{% endif %}
```
a-frazier, line 362: Creative-writing's `full`-level `userContent` has no `{% else %}` in this `{% case pov %}`, unlike the adjacent system template which always has one.

`pov` is read from `story.settings?.pov || 'second'` — a JSON blob with no runtime validation. If a story's stored `pov` is ever anything other than exactly `'first'`/`'second'`/`'third'` (legacy data, manual edit, future POV value), the entire `full` reinforcement text renders as an empty string — silently degrading to `none` with no error. The adventure system template's own POV branching always has a fallback, so this can't happen there.

src/lib/services/prompts/templates/narrative.ts
```diff
@@ -143,6 +160,23 @@ End with a natural opening for action, not a direct question.{% endif %}
 {% endif %}{% if tieredContextBlock != blank %}
 {{ tieredContextBlock }}
 {% endif %}{% if styleGuidance != blank %}{{ styleGuidance }}{% endif %}`,
+  userContent: `{%- case narratorReinforcement %}
+{%- when 'minimal' %}You are the narrator of this interactive adventure. I am the player controlling protagonist named {{ protagonistName }}.
+{%- when 'full' %}You are the narrator of this interactive adventure. I am the player controlling protagonist named {{ protagonistName }}. Write in {{ tense }} tense, {{pov}} person.
```
a-frazier: The adventure `full` `userContent` template leaks a stray literal space into the rendered turn message, from this nested `{% case pov %}` block (used only to set `actionExample`).

Verified by rendering directly with LiquidJS: for `narratorReinforcement='full'`, every `pov` produces `"...present tense, second person. \n\nYour role:"` — an extra space before the paragraph break, sent to the model on every adventure turn at the default reinforcement level. Cosmetic, but `narrative.test.ts` only asserts trimmed emptiness and absence of `{{`/`{%`, so nothing catches it.

a-frazier (commented):
Automated review findings (3 comments) — see inline. Highlights: a wasted per-turn render when a story has a custom system prompt, and a missing `{% else %}` fallback in the creative-writing `full` template that can silently blank the reinforcement text if `pov` is ever an unexpected value.

src/lib/services/ai/generation/NarrativeService.ts
collegatore: `buildPrompts` no longer renders the half it discards: `ContextBuilder` gains `renderTemplate` for a single id, `render` is expressed in terms of it, and the custom-prompt path now resolves and renders only `<templateId>-user`.

src/lib/services/prompts/templates/narrative.ts
collegatore: Added an `{% else %}` to both templates — third person, already the default elsewhere, is now the fallback branch. The adventure inner `case` that sets the action example had the same hole, so it got one too. Tests cover an unlisted pov still rendering `full` in both.

src/lib/services/prompts/templates/narrative.ts
collegatore: Fixed — the nested `case` now uses whitespace control on both sides, so no literal space survives. Added an assertion that `full` contains no space before a line break and no double space, for every pov and tense in both templates.

src/lib/services/prompts/templates/narrative.ts
```diff
@@ -143,6 +160,23 @@ End with a natural opening for action, not a direct question.{% endif %}
...
+{%- case pov -%}
+{%- when 'first' -%}{% assign actionExample = 'I push open the heavy door' %}
+{%- when 'second' -%}{% assign actionExample = 'You push open the heavy door' %}
+{%- else -%}{% assign actionExample = protagonistName | append: ' pushes open the heavy door' %}
```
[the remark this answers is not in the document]
collegatore: Fixed, but not by normalizing centrally: the two system halves fall back differently on purpose — adventure's pov chain ends in SECOND PERSON, creative-writing's in THIRD PERSON — so normalizing to one value would have silently moved creative-writing's fallback. Adventure's turn message now derives a `povWord` alongside the action example, so an unlisted pov reads second person throughout instead of naming the stored value back. Tests assert each template falls back the way its own system half does.

src/lib/services/prompts/templates/narratorReinforcement.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Any tag rather than a `{{ }}` match: the level is branched on, so it appears in
+ * `{% if %}`, `{% case %}` or `{% unless %}` as readily as in an output tag. Comments are
+ * removed and prose is ignored, so neither a branch someone commented out nor the variable's
+ * name written in the prompt text reports the setting as working.
+ */
+export function templateUsesNarratorReinforcement(content: string | null | undefined): boolean {
+  if (!content) return false
+  const active = content.replace(LIQUID_COMMENT, '')
+  const reference = new RegExp(`\\b${NARRATOR_REINFORCEMENT_VAR}\\b`)
+  return (active.match(LIQUID_EXPRESSION) ?? []).some((expression) => reference.test(expression))
```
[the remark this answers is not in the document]
collegatore: Both stripped before matching: `{% raw %}`/`{% endraw %}` joins comments as a region Liquid never evaluates, and string literals are removed from a tag's interior before the identifier is tested. A real reference can sit in neither, so no false negatives. Added coverage for raw blocks, `{% assign %}` with a quoted name, a quoted output, and a real branch alongside each.

PR #507, opened 2026-09-10
Bind a generation to the branch it started on

collegatore:
## The bug

Reported twice. Navigate between branches a few times without generating, then press Regenerate on the last message of the branch you land on. The branch is clipped by a dozen entries, the regeneration answers a prompt from a branch you left, and the Characters, Inventory and Story beats panels come up empty.

The panels are not merely blank — the rows have been deleted from the database and re-inserted under a different branch's id.

## Why

A retry snapshot recorded no branch. `RetryBackup` was keyed by story alone and `canRetry` compared only `storyId`, so a snapshot taken on branch A was offered, and restored, on branch B.

The damage follows from two things:

- **Positions are reused by sibling branches after a fork.** The rewind deletes `position >= entryCountBeforeAction`, so A's cut point lands mid-way through B.
- **`database.restoreRetryBackup` deletes the active branch's world-state rows and re-inserts the snapshot's**, which carry their own `branch_id`. The per-branch entity queries match `branch_id` exactly and never fall back to inherited rows, so B is left with none of its own.

That last part does not depend on the experimental settings: the legacy per-branch load and snapshot isolation both empty out, and main does too when the snapshot came from a branch. Lightweight branches only decide how total it is — a pre-snapshot COW branch still resolves its ancestors'.

## What changed

**Snapshots carry their branch.** `RetryBackup` and `PersistentRetryState` gain `branchId`; backups are keyed by story *and* branch, so each branch keeps its own and switching back restores the offer.

**Foreign snapshots are refused** in `RetryService.restoreFromBackup` — which is public and was reachable without any check — as well as in `preflight`, which keeps refusing earlier and with a better message. `restoreFromRetryBackup` asserts too, now that its payload carries a branch.

**Retry state persisted before this change is discarded on load.** It cannot be attributed to a branch, and guessing "main" is wrong for exactly the people who hit this bug. One lost cross-session retry per story; the next generation records an attributable snapshot.

**A generation and a branch switch are made mutually exclusive** by a lease held from before the snapshot is taken until after the last write settles. `ui.isGenerating` could not do this:

- it is set inside `generateResponse`, after the initiating handler has already taken a snapshot, translated input and written the user action;
- it is checked at `switchBranch`'s entrance, which a switch accepted while idle passes before a generation begins — `performBranchSwitch` awaits a database write before it assigns anything;
- Stop clears it without awaiting the pipeline, while a classification already entered keeps writing.

The lease is acquired by all four initiating handlers (`handleSubmit`, `handleRetryLastMessage`, `handleRegenerateNarration`, `handleRetry`) and passed into `generateResponse` as a required parameter, so a fifth caller cannot be added without one. Stop registers its rewind on the lease and waits for it rather than releasing, so the rewind cannot race the writes it exists to reverse.

**`addEntry` takes the branch to write to and throws on a mismatch.** Unreachable while the lease holds — that is the point. It is the tripwire if the exclusion is ever holed: a write lands as a thrown error rather than as a row on the wrong branch.

## The restriction, and when it can be lifted

Branch switching is refused for the duration of a generation. This stands in for something, and should not be removed on the grounds that it looks like unexplained caution.

`applyClassificationResult` reads the active branch and mutates the in-memory `characters`/`locations`/`items`/`storyBeats` arrays, which hold *that* branch's view. Pointing it at another branch means decoupling "the branch being written" from "the branch loaded in memory" — an architectural change, not a parameter. Redirecting only the entries would be worse than redirecting nothing: the narration would land on the branch that asked for it while the classification accounting for it did not.

**It can be lifted once world-state application takes the branch to write to as an argument instead of reading the active one, *and* `addEntry` redirects to the bound branch. Both, not either** — the entry half alone produces exactly the mismatch above. Branch-scoping `isGenerating` belongs to that work too.

This is recorded in `docs/architecture/overview.md` under "The generation lease", so it does not depend on this description surviving.

## Notes for review

- **No migration.** `stories.retry_state` is a JSON blob; `branchId` is added inside it, as `embeddedImageIds`, `characterSnapshots` and `timeTracker` were.
- **`BranchPanel.svelte` is mostly reindentation** from a new wrapper element — read it with `git diff -w` (23 insertions, 15 deletions of real change).
- The branch panel is covered and made `inert` rather than dimmed with a tooltip: on Android there is no cursor to change and no hover to read. Rename and delete become unreachable with it, which is accepted — both are decisions worth making against a branch you can still inspect.

## Testing

- `npm run check`, `npm test`, `npm run lint` all clean.
- 21 new unit tests: `RetryService.test.ts` covers both restore paths and direct entry through `restoreFromBackup`, that main and a named branch are not conflated, and that no state-mutating callback runs on refusal. `generationLease.test.ts` covers the drain/restore/release ordering, release after a throwing restore, and idempotent finish.
- **Manual verification of the concurrency orderings is still outstanding** — the delayed-switch, delayed-translation and delayed-classification cases, and the reported repro end to end. There is no DOM in the test environment, so the store and component work is only exercised by running the app.

## Summary by CodeRabbit

* **New Features**
  * Branch switching and cross-branch landmark navigation are blocked while story generation is active, with clear lock or error messaging.
  * Generation and retry actions remain tied to the story branch where they began.

* **Bug Fixes**
  * Prevented generation results, edits, deletions, and retry restores from affecting the wrong story or branch.
  * Retry options and checkpoint creation are unavailable when saved state is out of scope or generation is active.
  * Stopping or retrying generation now safely restores state after pending updates complete.

### DIFF
--- a/src/lib/components/story/ActionInput.svelte
+++ b/src/lib/components/story/ActionInput.svelte
@@ -469,6 +469,32 @@
     }
   }
 
+  /**
+   * Run a generation under its own lease.
+   *
+   * Acquiring is the first thing a handler does, before it touches anything: a refusal then
+   * has nothing to undo, and every side effect written above the acquire would otherwise
+   * inherit an obligation to reverse itself that nothing enforces.
+   *
+   * `generateResponse` takes the lease as a required parameter, so a handler cannot skip
+   * acquiring one. This is the other half — releasing — which a handler copied from another
+   * can lose without the compiler noticing.
+   */
+  async function withGenerationLease(run: (lease: GenerationLease) => Promise<void>) {
+    let lease: GenerationLease
+    try {
+      lease = story.acquireGenerationLease()
+    } catch (error) {
+      ui.showToast(errMessage(error), 'error')
+      return
+    }
+    try {
+      await run(lease)
+    } finally {
+      await lease.finish()
+    }
+  }
+
   async function generateResponse(
     lease: GenerationLease,
     userActionEntryId: string,
@@ -1026,41 +1052,28 @@
 
   async function handleSubmit() {
     if (!inputValue.trim() || ui.isGenerating || !story.currentStory) return
+    const currentStory = story.currentStory
 
-    ui.clearGenerationError()
-    ui.resetScrollBreak()
-    ui.clearSuggestions(story.currentStory.id)
+    await withGenerationLease(async (lease) => {
+      ui.clearGenerationError()
+      ui.resetScrollBreak()
+      ui.clearSuggestions(currentStory.id)
 
-    const rawInput = inputValue.trim()
-    const wasRawActionChoice = isRawActionChoice
-    const forceFreeMode = settings.uiSettings.disableActionPrefixes
+      const rawInput = inputValue.trim()
+      const wasRawActionChoice = isRawActionChoice
+      const forceFreeMode = settings.uiSettings.disableActionPrefixes
 
-    let content: string
-    if (isCreativeMode || wasRawActionChoice || forceFreeMode) content = rawInput
-    else content = actionPrefixes[actionType] + rawInput + actionSuffixes[actionType]
+      let content: string
+      if (isCreativeMode || wasRawActionChoice || forceFreeMode) content = rawInput
+      else content = actionPrefixes[actionType] + rawInput + actionSuffixes[actionType]
 
-    isRawActionChoice = false
-    inputValue = ''
-    if (textareaRef) textareaRef.scrollTop = 0
+      isRawActionChoice = false
+      inputValue = ''
+      if (textareaRef) textareaRef.scrollTop = 0
 
-    // Claimed before the snapshot: everything below reads or writes on this generation's
-    // behalf, and the awaits between here and the narration are long enough to switch in.
-    let lease: GenerationLease
-    try {
-      lease = story.acquireGenerationLease()
-    } catch (error) {
-      // The box was cleared above on the assumption this would go ahead. Give the text back
-      // rather than making the reader retype it.
-      inputValue = rawInput
-      isRawActionChoice = wasRawActionChoice
-      ui.showToast(errMessage(error), 'error')
-      return
-    }
-
-    try {
-      const embeddedImageIds = await database.getEmbeddedImageIdsForStory(story.currentStory.id)
+      const embeddedImageIds = await database.getEmbeddedImageIdsForStory(currentStory.id)
       ui.createRetryBackup(
-        story.currentStory.id,
+        currentStory.id,
         lease.branchId,
         story.entries,
         story.characters,
@@ -1072,7 +1085,7 @@
         rawInput,
         actionType,
         wasRawActionChoice,
-        story.currentStory.timeTracker,
+        currentStory.timeTracker,
       )
 
       const {
@@ -1096,9 +1109,7 @@
       // retrieval and classification working from a different wording than the narration —
       // and the retry path already passes `promptContent`, so the two disagreed.
       await generateResponse(lease, userActionEntry.id, promptContent, { inputTranslation })
-    } finally {
-      await lease.finish()
-    }
+    })
   }
 
   async function handleStopGeneration() {
@@ -1188,15 +1199,7 @@
     const error = ui.lastGenerationError
     if (!error || ui.isGenerating) return
 
-    let lease: GenerationLease
-    try {
-      lease = story.acquireGenerationLease()
-    } catch (err) {
-      ui.showToast(errMessage(err), 'error')
-      return
-    }
-
-    try {
+    await withGenerationLease(async (lease) => {
       const userActionEntry = story.entries.find((e) => e.id === error.userActionEntryId)
       if (!userActionEntry) {
         ui.clearGenerationError()
@@ -1219,9 +1222,7 @@
         countStyleReview: false,
         styleReviewSource: 'retry-error',
       })
-    } finally {
-      await lease.finish()
-    }
+    })
   }
 
   function dismissError() {
@@ -1248,15 +1249,7 @@
 
     // Claimed before the undo, which is itself an asynchronous write on this generation's
     // behalf.
-    let lease: GenerationLease
-    try {
-      lease = story.acquireGenerationLease()
-    } catch (error) {
-      ui.showToast(errMessage(error), 'error')
-      return
-    }
-
-    try {
+    await withGenerationLease(async (lease) => {
       let undo: { entitiesUndone: boolean; timeUndone: boolean }
       try {
         undo = await story.undoNarrationForRegenerate(entryId, lease)
@@ -1285,9 +1278,7 @@
         styleReviewSource: 'regenerate',
         cachedRetrievalResult: cachedRetrieval ? ui.retrievalResultFor(cachedRetrieval) : null,
       })
-    } finally {
-      await lease.finish()
-    }
+    })
   }
 
   async function handleRetryLastMessage() {
@@ -1306,15 +1297,7 @@
 
     // Claimed before the rewind, which deletes entries and entities on this generation's
     // behalf before the model is ever called.
-    let lease: GenerationLease
-    try {
-      lease = story.acquireGenerationLease()
-    } catch (error) {
-      ui.showToast(errMessage(error), 'error')
-      return
-    }
-
-    try {
+    await withGenerationLease(async (lease) => {
       const result = await retryService.handleRetryLastMessage(
         backup,
         {
@@ -1335,7 +1318,7 @@
           clearSuggestions: () => ui.clearSuggestions(storyId),
           clearActionChoices: () => ui.clearActionChoices(storyId),
         },
-        { storyId, branchId: story.currentStory.currentBranchId ?? null },
+        lease,
       )
 
       if (!result.success) {
@@ -1375,9 +1358,7 @@
       } finally {
         ui.setRetryingLastMessage(false)
       }
-    } finally {
-      await lease.finish()
-    }
+    })
   }
 
   function handleKeydown(event: KeyboardEvent) {
--- a/src/lib/components/story/ActionInput.svelte
+++ b/src/lib/components/story/ActionInput.svelte
@@ -1198,7 +1198,9 @@
       }
 
       try {
-        await story.deleteEntry(error.errorEntryId)
+        // The lease is this generation's own, so the guard lets it through: clearing the
+        // failed entry before regenerating is the holder tidying up after itself.
+        await story.deleteEntry(error.errorEntryId, lease)
       } catch (err) {
         // Regenerating over an entry that could not be removed would leave the failed one
         // above the new narration, so the retry stops here.
@@ -1251,7 +1253,7 @@
     try {
       let undo: { entitiesUndone: boolean; timeUndone: boolean }
       try {
-        undo = await story.undoNarrationForRegenerate(entryId)
+        undo = await story.undoNarrationForRegenerate(entryId, lease)
       } catch (error) {
         ui.showToast(errMessage(error), 'error')
         return
--- a/src/lib/components/story/ActionInput.svelte
+++ b/src/lib/components/story/ActionInput.svelte
@@ -56,8 +56,10 @@
     type BackgroundTaskInput,
     type PipelineUICallbacks,
     type PipelineEventState,
+    type RestoreResult,
   } from '$lib/services/generation'
   import { InlineImageTracker } from '$lib/services/ai/image'
+  import type { GenerationLease } from '$lib/utils/generationLease'
 
   function log(...args: any[]) {
     console.log('[ActionInput]', ...args)
@@ -468,6 +470,7 @@
   }
 
   async function generateResponse(
+    lease: GenerationLease,
     userActionEntryId: string,
     userActionContent: string,
     options?: {
@@ -509,6 +512,9 @@
     ui.startStreaming(visualProseMode, streamingEntryId)
 
     const currentStoryRef = story.currentStory
+    // The branch this generation is bound to. Read from the lease, not the live store: the
+    // store's value is what the lease exists to stop moving.
+    const leasedBranchId = lease.branchId
 
     let inlineImageTracker: InlineImageTracker | null = null
     if (inlineImageMode) {
@@ -681,7 +687,7 @@
             ) ?? undefined
           ui.setLastRetrievalResult(retrievalResult ?? null, {
             storyId: currentStoryRef.id,
-            branchId: currentStoryRef.currentBranchId ?? null,
+            branchId: leasedBranchId,
             position: storyPosition,
             actionContent: userActionContent,
           })
@@ -702,6 +708,7 @@
           narrationEntry = await story.addEntry(
             'narration',
             fullResponse,
+            leasedBranchId,
             generationMeta,
             fullReasoning || undefined,
             narrationEntryId,
@@ -794,7 +801,7 @@
 
       if (!fullResponse.trim()) {
         const errorMessage = 'The AI returned an empty response after 3 attempts. Please try again.'
-        const errorEntry = await story.addEntry('system', errorMessage)
+        const errorEntry = await story.addEntry('system', errorMessage, leasedBranchId)
         ui.setGenerationError({
           message: errorMessage,
           errorEntryId: errorEntry.id,
@@ -822,7 +829,7 @@
       // Deliberately not awaited — but the flag has to outlive the call, or the Memory
       // view will offer to create a chapter while this one is being created.
       const bgStoryId = currentStoryRef.id
-      const bgBranchId = currentStoryRef.currentBranchId ?? null
+      const bgBranchId = leasedBranchId
       ui.setBackgroundTasksActive(bgStoryId, bgBranchId, true)
       coordinator
         .runBackgroundTasks(input)
@@ -850,7 +857,11 @@
       const errorMessage = ui.wasBackgroundedDuringGeneration
         ? `Generation may have been interrupted while the app was in the background. ${baseMessage}`
         : baseMessage
-      const errorEntry = await story.addEntry('system', `Generation failed: ${errorMessage}`)
+      const errorEntry = await story.addEntry(
+        'system',
+        `Generation failed: ${errorMessage}`,
+        leasedBranchId,
+      )
       ui.setGenerationError({
         message: errorMessage,
         errorEntryId: errorEntry.id,
@@ -1023,43 +1034,58 @@
     inputValue = ''
     if (textareaRef) textareaRef.scrollTop = 0
 
-    const embeddedImageIds = await database.getEmbeddedImageIdsForStory(story.currentStory.id)
-    ui.createRetryBackup(
-      story.currentStory.id,
-      story.entries,
-      story.characters,
-      story.locations,
-      story.items,
-      story.storyBeats,
-      embeddedImageIds,
-      content,
-      rawInput,
-      actionType,
-      wasRawActionChoice,
-      story.currentStory.timeTracker,
-    )
-
-    const {
-      promptContent,
-      originalInput,
-      timing: inputTranslation,
-    } = await translateUserInput(content, settings.translationSettings)
-
-    const userActionEntry = await story.addEntry('user_action', promptContent)
-
-    if (originalInput) {
-      await database.updateStoryEntry(userActionEntry.id, { originalInput })
-      await story.refreshEntry(userActionEntry.id)
+    // Claimed before the snapshot: everything below reads or writes on this generation's
+    // behalf, and the awaits between here and the narration are long enough to switch in.
+    let lease: GenerationLease
+    try {
+      lease = story.acquireGenerationLease()
+    } catch (error) {
+      ui.showToast(errMessage(error), 'error')
+      return
     }
 
-    emitUserInput(content, isCreativeMode ? 'direction' : forceFreeMode ? 'free' : actionType)
-    await tick()
+    try {
+      const embeddedImageIds = await database.getEmbeddedImageIdsForStory(story.currentStory.id)
+      ui.createRetryBackup(
+        story.currentStory.id,
+        lease.branchId,
+        story.entries,
+        story.characters,
+        story.locations,
+        story.items,
+        story.storyBeats,
+        embeddedImageIds,
+        content,
+        rawInput,
+        actionType,
+        wasRawActionChoice,
+        story.currentStory.timeTracker,
+      )
 
-    // `promptContent`, not the raw `content`: with translation on the two differ, and the
-    // entry the narrator reads holds `promptContent`. Passing the raw text here left
-    // retrieval and classification working from a different wording than the narration —
-    // and the retry path already passes `promptContent`, so the two disagreed.
-    await generateResponse(userActionEntry.id, promptContent, { inputTranslation })
+      const {
+        promptContent,
+        originalInput,
+        timing: inputTranslation,
+      } = await translateUserInput(content, settings.translationSettings)
+
+      const userActionEntry = await story.addEntry('user_action', promptContent, lease.branchId)
+
+      if (originalInput) {
+        await database.updateStoryEntry(userActionEntry.id, { originalInput })
+        await story.refreshEntry(userActionEntry.id)
+      }
+
+      emitUserInput(content, isCreativeMode ? 'direction' : forceFreeMode ? 'free' : actionType)
+      await tick()
+
+      // `promptContent`, not the raw `content`: with translation on the two differ, and the
+      // entry the narrator reads holds `promptContent`. Passing the raw text here left
+      // retrieval and classification working from a different wording than the narration —
+      // and the retry path already passes `promptContent`, so the two disagreed.
+      await generateResponse(lease, userActionEntry.id, promptContent, { inputTranslation })
+    } finally {
+      await lease.finish()
+    }
   }
 
   async function handleStopGeneration() {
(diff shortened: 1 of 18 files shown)


src/lib/components/story/ActionInput.svelte
```diff
@@ -1023,43 +1034,58 @@
...
-      await story.refreshEntry(userActionEntry.id)
+    // Claimed before the snapshot: everything below reads or writes on this generation's
+    // behalf, and the awaits between here and the narration are long enough to switch in.
+    let lease: GenerationLease
+    try {
+      lease = story.acquireGenerationLease()
+    } catch (error) {
+      ui.showToast(errMessage(error), 'error')
+      return
     }
```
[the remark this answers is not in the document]
collegatore: Fixed in 1124ef42: `handleSubmit` now restores `inputValue` and `isRawActionChoice` before the toast, so a refused acquisition gives the typed text back instead of losing it.

src/lib/stores/story.svelte.ts
```diff
@@ -3691,6 +3703,15 @@ class StoryStore {
   ): Promise<Branch> {
     if (!this.currentStory) throw new Error('No story loaded')
 
+    // Creating a branch ends by switching to it, so it is a branch switch with a write in
+    // front. Refused here rather than at that switch, which would leave the branch created
+    // and the story still on the old one.
+    if (this.generationLease) {
```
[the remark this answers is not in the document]
collegatore: Fixed in 1124ef42, using the existing pending-switch counter rather than a new lock: `createBranchFromCheckpoint` now increments `pendingBranchSwitches` for the whole operation and releases it in a `finally`, and `acquireGenerationLease` already refuses while that is non-zero. The body moved to `performBranchCreation` so the bracket reads as one thing.

src/lib/stores/story.svelte.ts
```diff
@@ -3990,6 +4011,50 @@ class StoryStore {
...
+  acquireGenerationLease(): GenerationLease {
+    if (this.generationLease) {
+      throw new Error('A generation is already in progress')
+    }
+    if (this.pendingBranchSwitches > 0) {
+      throw new Error('A branch switch is still in progress')
+    }
+    if (!this.currentStory) throw new Error('No story loaded')
+
+    const lease = new GenerationLease(this.currentStory.currentBranchId ?? null, () => {
```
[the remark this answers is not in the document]
collegatore: Fixed in 1124ef42: `GenerationLease` now carries `storyId` alongside `branchId`, and `addEntry` validates both — two stories on main are both `null`, so the branch alone could not tell their writes apart. Story transitions are still not blocked outright; the write path is what corrupts, and the assertion turns a crossed write into a thrown error rather than a row on the wrong story.

src/lib/stores/ui.svelte.ts
```diff
@@ -839,18 +847,30 @@ class UIStore {
...
-    if (targetStoryId) {
-      this.retryBackups.delete(targetStoryId)
+    const scope = this.currentRetryScope
+    if (scope) {
+      this.retryBackups.delete(branchScopeKey(scope.storyId, scope.branchId))
 
       // Only clear from database if explicitly requested (user dismissed or used retry)
       if (clearFromDb) {
-        this.queueRetryStateWrite(() => database.clearRetryState(targetStoryId), 'clear')
+        this.queueRetryStateWrite(() => database.clearRetryState(scope.storyId), 'clear')
```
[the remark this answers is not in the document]
collegatore: Fixed in 1124ef42: the clear now reads the stored state first and only clears when its `branchId` matches the scope, so clearing branch A no longer drops branch B's persisted state from the single per-story slot.

src/lib/components/story/ActionInput.svelte
```diff
@@ -850,7 +857,7 @@
       const errorMessage = ui.wasBackgroundedDuringGeneration
         ? `Generation may have been interrupted while the app was in the background. ${baseMessage}`
         : baseMessage
-      const errorEntry = await story.addEntry('system', `Generation failed: ${errorMessage}`)
+      const errorEntry = await story.addEntry('system', `Generation failed: ${errorMessage}`, lease)
```
a-frazier: **High.** This `catch` handler's fallback is `story.addEntry('system', …, lease)` — the same call that just threw. Whenever `addEntry`'s new tripwire (or the pre-existing `'No story loaded'`) fires here, the handler throws again, so `ui.setGenerationError` and `notifyFailureIfBackgrounded` never run and the rejection escapes `generateResponse`/`handleSubmit` (neither has a `catch`; no global `unhandledrejection` handler).

Reachable today: `Header.goToLibrary` and the Android hardware-back handler both call `story.closeStory()` with no lease check, so pressing back mid-generation gives no toast, no system entry, no Retry affordance — just a silent console rejection. Line 804 (empty-response path) falls into the same `catch` and has the same shape.

The tripwire needs a fallback path that cannot itself trip.

src/lib/stores/story.svelte.ts
```diff
@@ -3990,6 +4033,54 @@ class StoryStore {
   private branchSwitchChain: Promise<unknown> = Promise.resolve()
   /** Sequence of the most recently requested switch; older queued ones are superseded. */
   private branchSwitchSeq = 0
+  /** Switches requested but not yet settled. A lease may not be taken against one. */
+  private pendingBranchSwitches = $state(0)
+
+  private generationLease = $state<GenerationLease | null>(null)
+
+  /** True while a generation holds the branch. Drives the switch affordances. */
+  get isGenerationLeaseHeld(): boolean {
```
a-frazier, line 4094: **Medium.** This tests only `this.generationLease !== null`, ignoring `lease.storyId` — the field `GenerationLease` carries expressly so "two stories' main branches are both `null`" can be told apart (the refusal at `acquireGenerationLease`, ~line 4141, has the same gap).

Concrete case: generate on story A, then press Android back / go to Library — `closeStory()` doesn't release or clear the lease — and open story B. B's Branches panel shows "A response is generating on this branch", `switchBranch` in B throws, and `acquireGenerationLease` in B refuses with "A generation is already in progress," for as long as A's classification takes to drain.

src/lib/components/story/ActionInput.svelte
```diff
@@ -822,7 +829,7 @@
       // Deliberately not awaited — but the flag has to outlive the call, or the Memory
       // view will offer to create a chapter while this one is being created.
       const bgStoryId = currentStoryRef.id
-      const bgBranchId = currentStoryRef.currentBranchId ?? null
+      const bgBranchId = leasedBranchId
       ui.setBackgroundTasksActive(bgStoryId, bgBranchId, true)
       coordinator
```
a-frazier, line 834: **Medium.** `coordinator.runBackgroundTasks(input)` is fire-and-forget, and the holder's `finally` (around the `handleSubmit` try block) releases the lease immediately after `generateResponse` returns — not after this settles.

`addChapter` (`story.svelte.ts:3122`) pushes into `this.chapters`, the in-memory list for whatever branch is loaded, and lore management writes lorebook entries. The branch panel unlocks the instant the lease releases — i.e. precisely while these background tasks are still running — so a switch made right then files the new chapter into the other branch's view.

`docs/architecture/overview.md`'s "The generation lease" section says the lease is held "after its last one [write]," which isn't true for these two. Either extend the lease over background tasks, or narrow that doc claim.

src/lib/utils/generationLease.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  /**
+   * Register work to run once the generation has drained, and get back a promise that
+   * settles when it has. The holder runs it; the caller only waits.
+   *
+   * Returns null if the generation has already finished, in which case there is nothing
+   * left to wait for and the caller should act directly.
+   */
+  deferRestore(restore: () => Promise<void>): Promise<void> | null {
```
a-frazier, line 40: **Low (latent).** A second `deferRestore` call silently replaces `restoreSettled` (line 43), so the first waiter's promise never settles and `await settled` in `handleStopGeneration` hangs forever.

Unreachable today only because `stopRequested` gates the single caller. In a class whose entire job is ordering, the safer shape is to return the existing promise (or throw) when a restore is already registered, rather than overwrite it silently.

src/lib/stores/ui.svelte.ts
```diff
@@ -839,18 +847,37 @@ class UIStore {
...
+      this.retryBackups.delete(branchScopeKey(scope.storyId, scope.branchId))
 
       // Only clear from database if explicitly requested (user dismissed or used retry)
       if (clearFromDb) {
-        this.queueRetryStateWrite(() => database.clearRetryState(targetStoryId), 'clear')
+        this.queueRetryStateWrite(async () => {
+          // One persisted slot per story, so clearing it for this branch would take whichever
+          // branch's state is actually in it. Clear only when it is this branch's.
+          const stored = (await database.getStory(scope.storyId))?.retryState
+          if (!stored || (stored.branchId ?? null) === scope.branchId) {
```
a-frazier, line 873: **Low.** A legacy persisted row (no `branchId`) can never be cleared while the scope is a named branch: `stored.branchId ?? null` is `null`, `scope.branchId` is an id, so this guard fails and the DB row survives.

Combined with `loadRetryBackupFromPersistent`'s discard-on-load (which also doesn't clear the row), such a row is re-read and re-discarded on every story load and never reclaimed. Worth clearing it at discard time instead.

src/lib/components/story/ActionInput.svelte
```diff
@@ -1079,27 +1105,50 @@
     ui.setLastLorebookRetrieval(null)
```
a-frazier: **Low.** `setLastLorebookRetrieval(null)` and `setLastRetrievalResult(null)` (this line and the next) run before the restore below is even attempted, so a refusal (foreign branch, or `assertEntriesRemovable` throwing inside `runRestore`) leaves them cleared with nothing undone — the exact failure mode `preflight`'s docstring says it exists to prevent. The backup itself survives, but lorebook stickiness and the retrieval cache are gone regardless.

src/lib/components/story/StoryEntry.svelte
```diff
@@ -174,11 +174,14 @@
...
 
   // Check if retry is available for this entry
+  // Branch as well as story: a snapshot taken elsewhere would be refused on restore, and
+  // offering it here hides the regenerate that does work on this branch.
   const canRetry = $derived(
     isLatestNarration &&
       ui.retryBackup &&
       story.currentStory &&
       ui.retryBackup.storyId === story.currentStory.id &&
+      (ui.retryBackup.branchId ?? null) === (story.currentStory.currentBranchId ?? null) &&
```
a-frazier, line 192: **Low.** `canSaveAndRegenerate` (further down in this file, ~line 382) didn't get this same branch check added. Harmless today only because `ui.retryBackup` is already scope-filtered by story+branch — which in turn makes this added check redundant, since a foreign-branch backup could never reach here in the first place. As it stands the two derivations now assert different invariants about the same value; either drop the redundant check here or add the equivalent one to `canSaveAndRegenerate` for consistency.

a-frazier (commented):
Reviewed the diff (`31e1c1bb..1124ef42`) with a specific trace through State Tracking, Rollback on Delete, and Lightweight Branches. The core fix is sound — `RetryBackup`/`PersistentRetryState` carrying `branchId`, branch-keyed backups, refusal in both `preflight` and the public `restoreFromBackup`, and the lease as real mutual exclusion rather than `ui.isGenerating`. Inline comments cover 7 findings on lines this diff touches; two more anchor to code the diff doesn't touch, so they're summarized here instead:

**`src/lib/services/database.ts:1726` — high, pre-existing, and the other half of the reported bug under Lightweight Branches.** `restoreRetryBackup` deletes world-state rows for the *current branch only*, but re-inserts every snapshot row under its own `branch_id` via a plain `INSERT` (`addCharacter`, `:934`). With `lightweightBranches` on and `snapshotComplete` false, `story.characters` comes from `getCharactersResolved` (`:1890`), which returns ancestor rows with their original `branch_id` — rows the delete never touches. Retry-last-message or Stop on a pre-snapshot COW branch then fails with `UNIQUE constraint failed: characters.id` *after* `deleteStoryEntries` and the four `DELETE`s have committed: the branch loses its entries and its own world state, exactly the damage this PR set out to stop. This PR's branch binding doesn't close it, because the branches match (restoring onto itself). Fix is to stamp snapshot rows to the target branch, or delete by resolved id set.

**`src/lib/stores/story.svelte.ts:1058` — medium, intersects State Tracking + Rollback on Delete.** `assertNotBusy` still checks only `_isRetryInProgress || ui.isGenerating` (mirrored by `entriesLocked` in `StoryEntry.svelte:212`), not the new lease — but the PR's own argument is that `ui.isGenerating` is set too late, and the lease now covers exactly that window. Two live windows where edit/delete stay enabled: (a) from `acquireGenerationLease` through `translateUserInput`/`addEntry` in `handleSubmit`, across the rewind in `handleRetryLastMessage`, and across `undoNarrationForRegenerate` in `handleRegenerateNarration`; (b) after Stop, `ActionInput.svelte:1097` clears `isGenerating` while `applyClassificationResult` is still writing and the deferred rewind hasn't begun. With State Tracking + Rollback on Delete, a delete in either window runs `rollbackFromPosition` + `deleteEntriesFromPosition` under the in-flight generation, so `entryCountBeforeAction` stops describing the story and the next Stop/Retry rewinds to the wrong cut point.

Checked and clean: all five `story.addEntry` call sites pass a lease with no stale positional args; `branchScopeKey`'s prefix scan is safe; `branchId` round-trips through save/load/import/export; `pendingBranchSwitches` increment/decrement can't deadlock; `GenerationLease.finish()` has no interleaving window when nothing was deferred; legacy and `snapshotComplete` COW branches load world state via `getCharactersForBranch`, confining the `database.ts` finding to pre-snapshot COW branches.

src/lib/components/story/ActionInput.svelte
collegatore: Fixed in 187b0173. Confirmed: the record and the throw were the same call. The error entry is now attempted in its own `try`, and a failure falls back to a toast so `setGenerationError`/`notifyFailureIfBackgrounded` still run instead of the rejection escaping. The empty-response path at :804 falls into the same handler and is covered by it.

src/lib/stores/story.svelte.ts
collegatore: Fixed in 187b0173. The refusals keep their scope — a lease from story A is still writing through this store, so letting B switch or generate would interleave — but they no longer misdescribe it: `generationBusyMessage` distinguishes the two cases, and the panel shows "A response in another story is still finishing" via `isGenerationLeaseForAnotherStory`. `closeStory` not releasing the lease is the root cause and is left as-is deliberately; the generation is still running.

src/lib/components/story/ActionInput.svelte
collegatore: Doc claim narrowed in 187b0173; the lease is deliberately not extended. The chapter-numbering consequence was traced earlier and ruled not worth the fix — the row keeps the correct `branch_id` (it comes from the frozen ref), only `number` is taken from the branch now loaded, and `currentBranchChapters` filters the stray out of the prompt path. "The generation lease" now states that background tasks outlive the lease, that lore management refuses a moved branch via `assertScope`, and that chapter creation does not.

src/lib/utils/generationLease.ts
collegatore: Fixed in 187b0173: `deferRestore` returns the already-registered promise instead of overwriting, so the first waiter can never be stranded. Added a test asserting the second caller gets the first's promise and the second restore never runs.

src/lib/stores/ui.svelte.ts
collegatore: Fixed in 187b0173: the row is cleared at discard time in `loadRetryBackupFromPersistent`, so it is no longer re-read and re-discarded on every load. The branch-aware clear could never match it, exactly as described.

src/lib/components/story/ActionInput.svelte
collegatore: Fixed in 187b0173: the eager `setLastLorebookRetrieval(null)` is gone — `RetryService` already clears it after preflight commits — and `setLastRetrievalResult(null)` moved to after `result.success`, so a refusal leaves both intact.

src/lib/components/story/StoryEntry.svelte
collegatore: Fixed in 187b0173 by adding the branch check to `canSaveAndRegenerate` rather than dropping it from `canRetry`. Both are redundant against the scope-filtered getter, as you note, but the file already carries the equally redundant `storyId` check in both — so matching that is more consistent than making one derivation leaner than the other.

src/lib/stores/story.svelte.ts
```diff
@@ -4423,6 +4578,16 @@ class StoryStore {
...
 
+    // Backstop for a caller that bypasses RetryService: this deletes the active branch's
+    // world state and re-inserts the snapshot's rows, which carry their own branch_id.
+    if ((backup.branchId ?? null) !== (this.currentStory.currentBranchId ?? null)) {
+      throw new Error('Cannot restore a retry backup taken on another branch')
+    }
+
+    // Captured here for the same reason addEntry captures its own: the rewind is many awaits
+    // long and a story load during it would point the database work at the wrong story.
+    const scope = { storyId: this.currentStory.id, branchId: backup.branchId ?? null }
```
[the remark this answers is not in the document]
collegatore: Fixed in e80ac4ea. Confirmed: the rewind is deferred until the generation drains and nothing refuses a story load in between, so two main branches both being `null` let it through. `RestoreScope { storyId, branchId }` now replaces the bare branch argument, compared in `refuseForeignScope` and again in `restoreFromRetryBackup`. `handleStopGeneration` reads the scope when the rewind runs rather than when Stop was pressed — capturing it would answer for the wrong moment. Two tests added for the cross-story case.

src/lib/components/story/StoryEntry.svelte
```diff
@@ -206,7 +209,11 @@
   )
 
   // A retry restore rewrites the same entries a generation does, and the store refuses both.
-  const entriesLocked = $derived(ui.isGenerating || story.isRetryInProgress)
+  // The lease, not just `isGenerating`: the store refuses on the same terms, and the flag is
+  // unset for the preparation before a turn and for the drain after Stop.
+  const entriesLocked = $derived(
+    ui.isGenerating || story.isRetryInProgress || story.isGenerationLeaseHeld,
```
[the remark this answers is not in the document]
collegatore: Fixed in e80ac4ea: `createCheckpoint` now calls `assertNotBusy('create a checkpoint')`, and `canCreateCheckpoint` includes `!entriesLocked` as proposed. The store guard is the one that matters, since `createBranchFromCheckpoint` copies a checkpoint wholesale — a snapshot of a half-written turn would propagate into every branch made from it.

src/lib/stores/story.svelte.ts
```diff
@@ -1039,15 +1087,29 @@ class StoryStore {
...
+   *
+   * `ui.isGenerating` is not enough on its own: it is set inside `generateResponse`, after
+   * the initiating handler has already taken a snapshot and written the user action, and it
+   * is cleared by Stop while the classification is still writing. The lease covers both of
+   * those, so an edit or a delete in either window can no longer slip past.
+   *
+   * `holder` is the generation's own lease, for the two paths that legitimately edit under
+   * one — the error retry's delete of the failed entry, and the regenerate's undo.
+   */
+  private assertNotBusy(action: string, holder?: GenerationLease): void {
```
a-frazier, line 1099: `assertNotBusy`'s holder bypass only covers the `generationLease` check, not the `ui.isGenerating` check, so a caller passing its own lease as `holder` is still rejected if `ui.isGenerating` happens to be true.

Today both call sites (`ActionInput.svelte`'s `handleRetry` and `handleRegenerateNarration`) guard with `if (ui.isGenerating) return` before acquiring the lease, so this never fires in practice. But the function's own contract ("holder is the generation's own lease... this refuses the ones that get through") implies the bypass should be uniform. Any future caller that invokes `deleteEntry`/`undoNarrationForRegenerate` with its own live lease while `ui.isGenerating` is still true (e.g. cleanup fired from inside `generateResponse` itself, before it flips `isGenerating` off) would get "Cannot ... while a generation or retry is in progress" even though it holds the lease that should legitimately authorize the write.

src/lib/services/database.ts
```diff
@@ -1730,23 +1730,38 @@ class DatabaseService {
 
     // Restore entries not necessary as we are only deleting redundant entries since backup
 
+    // The insert has to cover exactly what the delete above covered, and the delete is scoped
```
a-frazier: This adds an 11-line paragraph of design rationale inline instead of in `docs/architecture`, duplicating text just added to `docs/architecture/overview.md` in this same diff.

Repo `CLAUDE.md` says: "If the explanation genuinely needs a paragraph, it is design rationale: put it in the relevant `docs/architecture/*.md` and, at most, point at it." This diff does that correctly for `closeStory()`'s new JSDoc (which points at "The generation lease" section in `overview.md`), but this `restoreRetryBackup` comment restates the same branch-scoped-insert rationale that `overview.md`'s diff already spells out, with no pointer back — the two copies will drift the next time either is edited.

a-frazier (commented):
Follow-up review of commits since the last review pass, cross-referenced against existing CodeRabbit/collegatore threads to avoid duplicates.

src/lib/stores/story.svelte.ts
collegatore: Fixed in d012c649: the holder now clears the lease check and `ui.isGenerating` together, since both describe the same generation — testing them apart was the inconsistency. `_isRetryInProgress` still refuses everyone including the holder, because it means those entries are being rewritten right now, which is narrower than "a generation is running" and not the holder's to override. Also merged two docblocks that had stacked above the function.

src/lib/services/database.ts
collegatore: Fixed in d012c649: the paragraph moved into the retry-backup bullet in `docs/architecture/overview.md` and the comment is now two lines pointing there, matching what `closeStory`'s JSDoc already does. The doc gained the part worth keeping that the comment had and it did not — that the deletes commit before the insert fails, which is why it destroys rather than merely failing.

src/lib/services/generation/RetryService.ts
```diff
@@ -85,6 +89,12 @@ export interface RetryStoreCallbacks {
   setLastLorebookRetrieval: (result: null) => void
 }
 
+/** The story and branch a restore is allowed to touch. */
+export interface RestoreScope {
```
a-frazier: **Finding 1 — the branch-scope comparison is re-implemented six times, and the copies disagree** (reuse/altitude, from `docs/plans/pr507-simplify-followups.md`)

`RestoreScope` here is a field-for-field copy of `LoreCallbackScope` (`loreCallbacks.ts:30`), and the same "is this `{storyId, branchId}` the one that's open?" check is hand-rolled at six sites total: here (`RestoreScope`, `refuseForeignScope` below), `story.svelte.ts` (`addEntry`, `isOpen`, `restoreFromRetryBackup`), `StoryEntry.svelte`, and `ActionInput.svelte` (`activeScope()`). All of the PR's copies normalize with `(x.branchId ?? null) === (y.branchId ?? null)`, but the pre-existing `assertScope` in `loreCallbacks.ts:43` doesn't — so `undefined` and `null` mismatch there and match everywhere else, which is exactly the drift `branchScope.ts`'s docblock says must not happen.

Suggested fix: move a `BranchScope` type and `sameBranchScope` predicate into `src/lib/utils/branchScope.ts` beside `branchScopeKey`, alias `RestoreScope` to it, and call `sameBranchScope` from all six sites (this closes the `assertScope` normalization gap for free). Also expose `get currentScope(): BranchScope` on the store so `ActionInput.svelte:1121`'s `story.currentStory?.id ?? ''` fallback can go away.

src/lib/components/story/ActionInput.svelte
```diff
@@ -1023,43 +1043,62 @@
...
+    let lease: GenerationLease
+    try {
+      lease = story.acquireGenerationLease()
+    } catch (error) {
+      // The box was cleared above on the assumption this would go ahead. Give the text back
+      // rather than making the reader retype it.
+      inputValue = rawInput
+      isRawActionChoice = wasRawActionChoice
+      ui.showToast(errMessage(error), 'error')
+      return
```
a-frazier: **Finding 2 — `handleSubmit` mutates UI state before acquiring the lease, and the restore is incomplete** (altitude, from `docs/plans/pr507-simplify-followups.md`)

`clearGenerationError()`, `resetScrollBreak()` and `clearSuggestions()` (just above this hunk, lines 1030-1032) run before the lease is acquired and are never undone on refusal — only `inputValue` and `isRawActionChoice` are restored in the `catch` here. A refused submit silently destroys the reader's suggestions and error banner. Every side effect added above the acquire from here on inherits an obligation to reverse itself that nothing enforces.

Suggested fix: move `story.acquireGenerationLease()` to immediately after the initial `if (!inputValue.trim() || …) return` guard, before any mutation. The refusal branch becomes `ui.showToast(errMessage(error), 'error'); return` with nothing to undo, and the `inputValue = rawInput` restore (and its comment) disappear.

src/lib/components/story/ActionInput.svelte
```diff
@@ -1023,43 +1043,62 @@
...
-      originalInput,
-      timing: inputTranslation,
-    } = await translateUserInput(content, settings.translationSettings)
-
-    const userActionEntry = await story.addEntry('user_action', promptContent)
-
-    if (originalInput) {
-      await database.updateStoryEntry(userActionEntry.id, { originalInput })
-      await story.refreshEntry(userActionEntry.id)
+    // Claimed before the snapshot: everything below reads or writes on this generation's
```
a-frazier: **Finding 3 — the lease acquire/release scaffold is pasted into all four initiating handlers** (simplification, from `docs/plans/pr507-simplify-followups.md`)

This acquire / toast-on-refusal / return, then `try … finally { await lease.finish() }` scaffold is repeated verbatim here and at `:1191`, `:1250`, `:1308` (`handleRetryLastMessage`, `handleRegenerateNarration`, `handleRetry`). A fifth handler can't skip acquiring a lease — it's a required parameter — but nothing makes the *release* half hard to get wrong, since a new handler tends to get written by copying whichever of the four the author lands on.

Suggested fix — one local helper beside `generateResponse`:
```ts
async function withGenerationLease(
  run: (lease: GenerationLease) => Promise<void>,
  onRefused?: () => void,
): Promise<void> {
  let lease: GenerationLease
  try {
    lease = story.acquireGenerationLease()
  } catch (error) {
    onRefused?.()
    ui.showToast(errMessage(error), 'error')
    return
  }
  try {
    await run(lease)
  } finally {
    await lease.finish()
  }
}
```
Each handler becomes `await withGenerationLease(async (lease) => { … })`; Finding 2 removes the need for `onRefused` in `handleSubmit`.

a-frazier (commented):
First three findings from the `/simplify` follow-up handover (`docs/plans/pr507-simplify-followups.md`), posted as inline comments.

src/lib/services/generation/RetryService.ts
collegatore: Done in 8bf8ced5: `BranchScope` and `sameBranchScope` sit beside `branchScopeKey`, `RestoreScope` and `LoreCallbackScope` are aliases of it, and all six sites call the predicate. The store gained `currentScope` so nothing outside it assembles the pair by hand. The `assertScope` normalisation gap closed with it, as you said it would.

src/lib/components/story/ActionInput.svelte
collegatore: Done in 8bf8ced5: the acquire is now the first thing `handleSubmit` does after its guard, so `clearGenerationError`/`resetScrollBreak`/`clearSuggestions` only run once the lease is held. The refusal path has nothing to undo — the `inputValue = rawInput` restore and its comment are gone. That restore was itself a fix from an earlier round, which rather makes your point about the obligation nothing enforces.

src/lib/components/story/ActionInput.svelte
collegatore: Done in 8bf8ced5: `withGenerationLease` holds the scaffold and all four handlers use it; there is one `acquireGenerationLease()` call in the file now. Took it without `onRefused`, since Finding 2 removed the only caller that needed it. `handleRetryLastMessage` also passes the lease itself as the restore scope rather than rebuilding one from the live store — it runs under the lease, so the lease is what it should target.

a-frazier (approved):

PR #516, opened 2026-09-15
Remove automatic pre-sync backup checkpoints

collegatore:
## Summary

- remove automatic `Pre-sync backup` checkpoint creation from push, pull replacement, and received-push replacement flows
- preserve transfer of checkpoints that already belong to the source story
- centralize sync push/import orchestration and add regression coverage for replacement and fresh imports
- make the receiving modal show accurate preparing, importing, conflict-retry, and success states without exposing stale actions

## Testing

- `npm test` — 1,647 passed, 9 skipped
- `npm run check` — 0 errors, 0 warnings
- `npm run lint -- --quiet` — 0 errors
- `git diff --check upstream/master...HEAD`

## Summary by CodeRabbit

* **Improvements**
  * Streamlined story synchronization when receiving, pulling, and pushing stories.
  * Added clearer progress indicators and status messaging during imports.
  * Improved conflict and replacement messaging, including guidance when pack selection is required.
  * Added retry support after failed imports and cancelled pack selection.

* **Bug Fixes**
  * Preserved story checkpoints when exporting stories for synchronization.
  * Improved pack-binding handling and import error recovery.
  * Prevented stale sync sessions from overwriting current story data.


collegatore:
Addressed the pack-resolution rejection path in `86a21af6`.

`resolveIncomingPack()` now has its own rejection boundary. On failure, the modal reports the error and clears `receivingStory`, while retaining the received JSON and preview so the user can retry instead of being stuck until reset.

Validation:
- focused sync tests: 11 passed
- `npm run check`: 0 errors, 0 warnings
- changed-file ESLint: 0 errors
- pre-push check and repository lint completed successfully

collegatore:
This [comment](https://github.com/AventurasTeam/Aventuras/pull/516#pullrequestreview-5213656885) is SO MUCH out of scope; I won't even start on it. We have had this bug for aeons, and it must not stop anyone from fixing adjacent failures.

src/lib/components/sync/SyncModal.svelte
```diff
@@ -226,16 +251,8 @@
     showReceivedConflict = false
 
     try {
-      // If replacing, delete the existing story first
       const existingId = await syncService.findStoryIdByTitle(receivedStoryPreview.title)
```
a-frazier: `importReceivedStory` reads live `receivedStoryJson`/`receivedStoryPreview` state instead of a captured snapshot, and closing/reopening the modal can null that state out from under an in-flight resolution.

**Failure scenario:** User receives a story while `resolveIncomingPack` is auto-resolving (no dialog shown, so `packMapping` stays null). User closes and reopens the sync modal; `close()`'s `cancelPendingPackMapping()` is a no-op since `packMapping` is still null, so the pending resolution isn't cancelled. `resetState()` nulls `receivedStoryPreview`/`receivedStoryJson` while the old promise is still pending. It later resolves and hits `receivedStoryPreview.title` on null → TypeError, caught by the outer catch as a confusing "Cannot read properties of null" error, and the shared `finally` force-resets `loading`/`receivingStory`/`receivedStoryJson`/`receivedStoryPreview` — potentially clobbering a new session started in the reopened modal.

src/lib/services/syncActions.ts
```diff
@@ -0,0 +1,24 @@
...
+  const storyJson = await syncService.exportStoryToJson(storyId)
+  await syncService.pushStory(connection, storyJson)
+}
+
+export async function importSyncedStory(
+  storyJson: string,
+  existingStoryId: string | null,
+  packBinding: PackBindingResolution,
+): Promise<ImportResult> {
+  if (existingStoryId) await syncService.deleteStory(existingStoryId)
```
a-frazier, line 19: `importSyncedStory` deletes the existing local story before the replacement import is confirmed to succeed, with no rollback, and the automatic backup checkpoint that used to run here has been removed entirely.

**Failure scenario:** User pulls or receives a story that replaces an existing local one. `deleteStory(existingStoryId)` runs, then `exportService.importFromContent` throws or returns `{success:false}` (malformed payload, DB error, pack-write failure). The original story is now gone and nothing replaces it — `SyncModal` only surfaces `error = result.error`, with no recovery path.

(Note: verified the removed "Pre-sync backup" checkpoint wouldn't have protected against this either — it was destroyed by `ON DELETE CASCADE` the moment `deleteStory` ran, even before this PR. So this is a real, still-open risk, not a new regression this diff introduces — but worth addressing given the checkpoint safety net is now gone entirely.)

src/lib/components/sync/SyncModal.svelte
```diff
@@ -184,11 +186,21 @@
...
+    try {
+      const exists = await syncService.checkStoryExists(preview.title)
+      if (exists) {
+        receivingStory = false
+        showReceivedConflict = true
+      } else {
+        receivingStory = false
+        await importReceivedStory()
+      }
+    } catch (e) {
```
a-frazier, line 209: This catch block sets `error`/`receivingStory` but never calls `resumeReceivedStories()`, unlike every sibling exit path in the file (the invalid-payload branch, `discardReceivedStory`, `importReceivedStory`'s `finally`).

**Failure scenario:** `syncService.checkStoryExists(preview.title)` throws (e.g. a transient Tauri IPC/DB error). Polling was already stopped before `claimReceivedStory` was invoked and is never restarted automatically; queued/incoming stories sit idle until the user manually clicks Discard or retries Import on the stuck preview.

a-frazier (commented):
Posting the 3 correctness findings from an automated code review as inline comments.

src/lib/components/sync/SyncModal.svelte
collegatore: Fixed in `74c52827`.

`importReceivedStory` now captures the preview alongside the payload and uses both throughout, stops before the delete+import if the pending payload is no longer the one it started with, and replaces the blanket `finally` with an identity-guarded teardown — so the error, the success message, the payload clearing and `resumeReceivedStories()` only touch the session the import belongs to.

Validation: full suite 1,647 passed / 9 skipped, `npm run check` 0 errors 0 warnings, ESLint and Prettier clean on the changed file.

src/lib/services/syncActions.ts
collegatore: Not taking this one here, for the reason you identified yourself: the deleted story was already unrecoverable via `ON DELETE CASCADE` before this PR, so it is a long-standing gap rather than a regression this diff introduces. Same finding as the earlier CodeRabbit review on this PR, and the same answer — it must not hold up the adjacent failures being fixed.

It is worth doing properly (snapshot-and-restore, or import-to-temp-then-swap), which is its own change, not a rider on this one.

src/lib/components/sync/SyncModal.svelte
collegatore: I don't think this one is a bug. The sibling paths that call `resumeReceivedStories()` — the invalid-payload branch, `discardReceivedStory`, and `importReceivedStory`'s teardown — are exactly the ones that *drop* the payload. This catch keeps it pending, and the UI renders the "Try importing again, or discard the story" card with both buttons, each of which resumes.

Resuming here would be actively wrong: with a non-empty queue it would claim the next story and overwrite the preview the user was just asked about. Polling staying stopped while a story is pending is the same behaviour as every other pending state.

src/lib/components/sync/SyncModal.svelte
```diff
@@ -204,60 +216,81 @@
...
+      // user-cancelled choice as retryable when the same received story is still active.
+      if (receivedStoryJson === pendingStoryJson) receivedStoryNeedsPack = true
       receivingStory = false
       return
     }
 
+    // Everything below writes to the database, so a resolution that arrives after the modal was
+    // reset — or after a newer payload took over — must stop before it replaces a story the user
+    // is no longer talking about.
+    if (receivedStoryJson !== pendingStoryJson) return
```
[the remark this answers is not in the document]
collegatore: Done in `1f2d7324` — replaced both payload-identity guards with a monotonic `receivedSession` counter, bumped on claim, discard, and reset, and captured per import.

src/lib/components/sync/SyncModal.svelte
```diff
@@ -204,60 +216,81 @@
...
+      imported = result.success
+      if (!result.success) importError = result.error ?? 'Import failed'
+    } catch (e) {
+      importError = e instanceof Error ? e.message : 'Import failed'
+    }
 
-      const result = await exportService.importFromContent(receivedStoryJson, true, {
-        resolvePackBinding: async () => packBinding,
-      })
+    if (imported) await story.loadAllStories()
```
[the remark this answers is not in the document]
collegatore: Done in `1f2d7324` — `story.loadAllStories()` now runs inside the guarded block with its own catch, so a reload failure falls through to the teardown without reporting a successful import as failed.

src/lib/components/sync/SyncModal.svelte
```diff
@@ -204,60 +216,81 @@
...
+    if (receivedStoryJson !== pendingStoryJson) return
+
+    loading = false
+    receivingStory = false
+    receivedStoryJson = null
+    receivedStoryPreview = null
+    if (importError) {
+      error = importError
+    } else {
+      syncSuccess = true
```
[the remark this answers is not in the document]
collegatore: Done in `1f2d7324` — the payload is now kept on failure and the queue resumes only after success; restoring the deleted local story stays out of scope for this PR.

a-frazier (approved):

PR #517, opened 2026-09-16
Derive checkpoint ownership from the entries table

collegatore:
## Why

A checkpoint has no branch column. It is owned by the branch of the entry it is anchored to, and that was read out of `entriesSnapshot` — a copy of the whole story up to that point, held for every checkpoint in the open story for the whole session.

Measured on a real save: the worst story parsed **12.56 MB of JSON into 22.77 MB of heap**, 6141 entry copies of 1584 live entries, 5.8x the story's own text. Everything read from it was one `branchId`.

The same derivation could not tell a checkpoint whose entry was deleted from one that simply belongs to another branch — the in-memory entry list holds only the current lineage, so both vanish from the landmark list. A checkpoint stranded by a rollback was invisible and unremovable.

## What changed

- **Ownership resolves in SQL**, through a LEFT JOIN on `story_entries`, surfacing as `branchId` plus `anchored`. Left, not inner: dropping the unanchored rows is the behaviour being removed. Three copies of the old helper are gone.
- **The store stops loading `entries_snapshot`.** `Checkpoint` (loaded) and `CheckpointRecord` (persisted) split the shapes; export, sync and backup keep the whole row, so the file format and the import round-trip are unchanged.
- **Branch cleanup derives its own set**, as a subquery inside the deletion transaction, ahead of the statement deleting the entries it reads. It no longer depends on the store's list being complete. This is the rule `persistence.md` already states for runtime variables — *"a read the batch depends on belongs inside it"* — applied to branch deletion.
- **Orphaned checkpoints are listed**, in a foldable section below the landmarks that appears only when the story has any, folded on open, counted in its header, with `orphaned` where the branch name sits. Rows are inert; rename and delete still work, so the panel is where they get cleared out.
- **`database.restoreCheckpoint` is removed** — unreachable since restoration was replaced by branching, and its last tie was the field this drops.

No schema change, no migration.

## Result

| Story | Checkpoints | Dropped | Retained |
| --- | ---: | ---: | ---: |
| S1 | 24 | 12.56 MB | 0.02 MB |
| S2 | 13 | 3.37 MB | 0.13 MB |
| S3 | 9 | 2.24 MB | 7.38 MB |

Worst story: **22.82 MB of heap → 0.04 MB**. That 100% is flattering and story-specific — S3 keeps 7.38 MB of the character/location/item/beat/lorebook snapshots, because branch creation genuinely seeds a new branch from those. The claim that generalises is narrower: the entry snapshot is no longer loaded at all.

## Worth a reviewer's attention

The type split earned itself during implementation. `export.ts` turned out to be a **third** export path nobody had catalogued; typing the payload as `CheckpointRecord[]` made it a compile error instead of an export silently shipping without the snapshots `import/structure.ts` maps unguarded — which would have surfaced as `undefined.map` on someone else's re-import.

## Testing

`npm run check` 0 errors · `npm test` 1667 passed, 0 failed · `npm run lint` 0 errors (232 pre-existing `boundaries/dependencies` warnings).

New persistence tests cover an anchor on a branch, on main, and missing, and that the orphan survives the read. Branch-deletion tests assert the statement set, that no checkpoint ids come from the caller, and that the checkpoint delete precedes the entries delete it depends on. Landmark tests keep the two kinds of miss apart.

Verified in the running app against an import fixture carrying two orphaned checkpoints and five branches — there is no DOM in the test environment, so the panel markup cannot be covered by the suite.

## Known, pre-existing, not fixed here

Deleting a branch can still clear a surviving branch's fork checkpoint reference, where that branch was created from a checkpoint anchored in another branch. `branches.checkpoint_id` is `ON DELETE SET NULL`, so it goes silently. CodeRabbit raised this on #515 and it merged unaddressed.

Its reachability argument there was wrong — branch creation cannot produce the pairing, because both entry points constrain the checkpoint to the lineage being read and the child-branch guard then holds. **Import can**: `import/structure.ts` writes `parentBranchId` and `checkpointId` from the file's graph without checking they agree. The defect is real, by that door. Closing it means deciding between refusing the deletion and preserving the checkpoint, which is a behaviour question this change deliberately leaves open.

## Summary by CodeRabbit

- **New Features**
  - Added a collapsible **Orphaned** section to story navigation for checkpoints whose entries are no longer available.
  - Orphaned checkpoints can be renamed or deleted and are clearly identified on touch devices.

- **Bug Fixes**
  - Improved checkpoint ownership and branch association, including handling of missing or deleted entries.
  - Fixed branch deletion so associated checkpoints are cleaned up reliably.
  - Improved checkpoint preservation across exports, imports, and synchronization.


src/lib/components/layout/StoryNavPanel.svelte
```diff
@@ -38,9 +40,15 @@
...
-  const landmarks = $derived(
+  const landmarkList = $derived(
     buildLandmarks(story.entries, story.checkpoints, story.branches, activeBranch),
   )
+  const landmarks = $derived(landmarkList.landmarks)
+  const orphaned = $derived(landmarkList.orphaned)
+
+  // Not persisted with the panel's own state: a reader who opens this to clear one checkpoint out
+  // does not want it open on every story afterwards.
+  let orphansExpanded = $state(false)
```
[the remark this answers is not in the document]
collegatore: Confirmed and fixed in c88709e7. `runStoryLoad` reassigns `currentStory` rather than clearing it, and only `resetStoryState` nulls it, so AppShell's `{#if ui.navPanelOpen && story.currentStory}` never goes false on a story switch and the component keeps its state. Added a `$effect` resetting `orphansExpanded` when the story id changes, and made the requirement say so rather than leaving it to an unmount the component does not control.

collegatore:
On the outside-diff finding about `sync.ts` / `importSyncedStory` — it's valid, and it isn't this PR's code.

`services/syncActions.ts` was added by #516 (`01c46144`). This branch sees it only because it merged `master` in to resolve a semantic conflict: #516 added a third `db.getCheckpoints` mock that collided with the rename here. `git diff upstream/master...HEAD` doesn't list `syncActions.ts` at all, and this PR's entire footprint in `sync.ts` is one line — `getCheckpoints` → `getCheckpointRecords`.

The mechanics do check out: `importSyncedStory` calls `syncService.deleteStory` → `database.deleteStory`, `stories.id` cascades to entries, chapters and checkpoints, and that commits before `importFromContent` is attempted. A failure after it leaves the old story gone with only the partial new one rolled back. Worth noting the removed `createPreSyncBackup` wasn't protecting against this either — its checkpoint belonged to the story being deleted and `checkpoints.story_id` is `ON DELETE CASCADE`, so it died with the row it was meant to save.

Not fixing it here: it's a Major data-integrity change to the replacement flow, and burying it in a checkpoint-ownership PR serves nobody. It's tracked on our side for whoever owns the sync path. The shape to reach for is import-then-swap rather than delete-then-import.

src/lib/components/branch/BranchPanel.svelte
```diff
@@ -163,7 +156,7 @@
   const createBranchTitle = $derived.by(() => {
     if (canCreateBranch) return 'Create new branch from latest checkpoint'
     const currentBranchId = story.currentStory?.currentBranchId ?? null
-    const hadOne = story.checkpoints.some((c) => getCheckpointBranchId(c) === currentBranchId)
+    const hadOne = story.checkpoints.some((c) => c.branchId === currentBranchId)
```
a-frazier: **Correctness:** `createBranchTitle`'s `hadOne` check compares `c.branchId === currentBranchId`, but an orphaned checkpoint always has `branchId: null` regardless of the branch it was actually created on, so it collides with the main branch (`currentBranchId === null`).

Failure scenario: a checkpoint created on branch 'B' has its anchoring entry deleted (rollback), leaving it orphaned (`branchId` now `null`, `anchored: false`). Switch to main with zero real main-branch checkpoints — `hadOne` evaluates `true` because the orphaned checkpoint's `null` branchId matches main's `null` `currentBranchId`, so the disabled button shows "This branch's checkpoints aren't usable..." instead of the correct "No checkpoints available...".

This is a regression: the pre-PR code read the branch id out of the checkpoint's own frozen `entriesSnapshot`, which preserved the true creation branch even after the live entry was deleted, so this collision is new.

src/lib/utils/storyNavigation.ts
```diff
@@ -116,7 +128,18 @@ export function buildLandmarks(
     }
   }
 
+  const orphaned: (OrphanedCheckpoint & { createdAt: number })[] = []
+
   for (const checkpoint of checkpoints) {
+    if (!checkpoint.anchored) {
+      orphaned.push({
```
a-frazier, line 155: **Correctness:** the `!checkpoint.anchored` branch pushes to `orphaned` unconditionally, without excluding a checkpoint that is also `activeBranch.checkpointId` and already rendered as the 'origin' landmark a few lines above.

Failure scenario: import data can make `branch.checkpointId` disagree with `branch.forkEntryId` (acknowledged in the PR description as a real, open gap: "import writes parentBranchId and checkpointId... without checking they agree"). If the origin checkpoint for the active branch is itself unanchored while `forkEntry` still resolves in the loaded lineage, the origin landmark is pushed using that checkpoint's name, and the same checkpoint is separately pushed into `orphaned` here with no de-dup against `activeBranch?.checkpointId` — the checkpoint appears twice in the nav panel (once as the branch origin, once under "Orphaned").

Narrower edge case than a typical bug since it depends on the already-flagged import inconsistency, but worth a guard given the PR explicitly calls that gap out as reachable.

a-frazier (commented):
Two correctness findings from an automated review of the checkpoint-ownership rewrite.

src/lib/components/branch/BranchPanel.svelte
collegatore: Confirmed, and it is a regression exactly as described — the old lookup read the frozen `entriesSnapshot`, which kept the creation branch after the entry went, while the join now yields `branchId: null` for any orphan. Fixed in 20cf801f. `getLatestCheckpoint` was saved by its `liveEntryIds` guard; `hadOne` had none. Both call sites now go through a shared `checkpointsOnBranch` helper in `storyNavigation.ts` that applies `anchored` once — in a plain module, so the suite can cover it, which a `.svelte` file cannot be.

src/lib/utils/storyNavigation.ts
collegatore: Confirmed and fixed in 20cf801f. The anchored path already skipped `activeBranch.checkpointId`; the unanchored path did not, so the same checkpoint could be both the origin row and an orphan row. It is now skipped only when an origin row was actually rendered for it — where the fork entry is missing there is no origin row to hold it, and dropping it there would hide the orphan the section exists to surface. Both cases are covered by new `buildLandmarks` tests.

PR #520, opened 2026-09-17
Show entry numbers in the memory panel and the story view

collegatore:
## Why

Entry numbers already exist and are already how a reader names a place in the story — the navigation panel jumps to one, the timeline panel labels its findings with one, a narration's Response info reports its own. But the number is only ever visible where one of those surfaces happens to print it. Reading the story itself, a reader cannot see where they are; and the Memory panel, which is entirely about which stretch of story a chapter covers, never named that stretch — a chapter card said `26 entries` and its expanded list showed untitled excerpts with no way back to the text they came from.

## What changed

**Memory panel**

- A chapter card reports the range of entries it covers beside the count, marked with a milestone glyph: `26 entries (3 → 29)`. The two figures are derived independently — on a story whose numbering has gaps they disagree, and both are true, so neither is adjusted to match the other. A one-entry chapter names its entry once rather than as a range to itself.
- Each row of an expanded chapter shows its entry number, and the row itself jumps to that entry. The list says so above it.
- The ten-row cap on that list is gone. It hid entries the card's new range advertises.

**Story view**

- A new Appearance setting, *Show Entry Number and Time*, off by default, puts a narration's number and the in-story time it covers on the entry, each marked by a glyph rather than a label. The number keeps a fixed place on the header at every width; the time sits there where there is room and moves to a row of its own where there is not.
- Player actions and system entries are excluded: an action is an instant with no span to show, and a system entry is an application message rather than story. Where a narration records no time, it says so rather than showing nothing.

**Entry header sizing** — a fix found along the way

The header hid its chips and folded its toolbar on viewport breakpoints, which could never fire on desktop: `tauri.conf.json` floors the window at 800px and the gate sat at 640px. Worse, the thing that actually narrows the story column is a panel opening beside it, which a media query cannot see at all. The entry card is now a query container and its header gives way against its own width, in order: the in-story time drops to its own row, then the turn-duration chip goes, then the token count, then the buttons fold into the overflow menu.

## Notable

- **The duration chip and its fallback are decided by one measurement.** The chip toggles the generation-activity timeline, and its fallback is a toggle in Response info — which renders in a portal outside the card, where no container query can reach. Gating the two separately left a band of widths where the chip had gone and the overflow menu had not yet arrived, so a report auto-opened during generation could not be closed. One measured card width now decides both, so exactly one of them exists at any width.
- `jumpToEntry` consolidates the guard-request-switch-confirm sequence that `StoryNavPanel` and `BranchPanel` had each inlined. They had already drifted: one closed the nav panel on mobile and the other the sidebar; one reported "not in this branch" on a miss and the other reported nothing.
- `retry` entries are treated as narrations. No code path creates one any more, and every other reader in the tree already treats an old `retry` entry as a narration.

## Testing

`npm run check`, `npm test` and `npm run lint` all pass; 11 new unit tests cover `jumpToEntry` and `entryNumberRange`, including the gapped-numbering case.

Not yet exercised: the narrowest header grades need a card under ~27rem, which an 800px window floor cannot produce — those want the Android emulator. The container also establishes a containing block for fixed-position descendants; the Response info popover portals to the document body and so escapes it, but the overflow menu is worth a look in a running app.

## Summary by CodeRabbit

* **New Features**
  * Added an option to show entry numbers and in-story times on story entries.
  * Chapter cards now display entry counts and entry number ranges.
  * Chapter entry lists now show all entries with numbered navigation buttons.
  * Improved navigation to entries, including mobile panel handling and clearer confirmations.

* **UI Improvements**
  * Updated entry layouts to adapt across screen sizes.
  * Activity details remain accessible when space is limited.


src/lib/components/story/StoryEntry.svelte
```diff
@@ -1692,6 +1714,22 @@
...
+    <span class="bg-muted rounded px-1.5 py-0.5 text-[11px] tabular-nums">
+      Story time: {generationInfo.storyTime || 'not recorded'}
+    </span>
+  {/snippet}
+
+  {#if showEntryMeta}
+    <div class="text-muted-foreground mb-2 flex items-center @min-[37.6rem]:hidden">
+      {@render entryMeta()}
+    </div>
+  {/if}
```
a-frazier, line 1755: This information is already available in the Response Info popover defined starting at Line 1511 in your branch, and story entries are already fairly busy UI wise.  Do we really need this?

src/lib/components/layout/StoryNavPanel.svelte
```diff
@@ -68,22 +69,14 @@
...
-      }
-    }
+    jumpToEntry({
+      entries: story.entries,
+      entryId,
+      ui,
+      confirmation,
+      closeOnMobile: () => ui.closeNavPanelOnMobile(),
+      canHover: supportsHover(),
+    })
```
a-frazier: Since this is the entire function body for `goTo`, we might as well inline it.

src/lib/components/memory/ChapterEntryList.svelte
```diff
@@ -31,35 +35,56 @@
...
+      canHover: supportsHover(),
+    })
+  }
 </script>
 
 {#if expanded && entries.length > 0}
   <div class="mt-2 space-y-1 border-l-2 pl-2" transition:slide={{ duration: 200 }}>
-    {#each entries.slice(0, 10) as entry (entry.id)}
+    <!-- Uncapped: the card advertises a range, so every entry in it has to be reachable. -->
+    {#each entries as entry (entry.id)}
```
a-frazier, line 61: ❤️

src/lib/components/story/StoryEntry.svelte
```diff
@@ -1375,12 +1385,13 @@
...
 
-    <!-- How long the turn took, opening its timeline. Hidden on narrow screens for the same
+    <!-- How long the turn took, opening its timeline. Hidden on a narrow card for the same
          reason as the token badge above; the overflow menu carries it there instead. -->
     {#if activityRecord}
+      <Separator orientation="vertical" class="hidden h-4 @min-[37.6rem]:block" />
       <button
         type="button"
-        class="bg-muted text-muted-foreground hover:text-foreground hidden rounded px-1.5 py-0.5 text-[11px] tabular-nums transition-colors sm:inline"
+        class="bg-muted text-muted-foreground hover:text-foreground hidden rounded px-1.5 py-0.5 text-[11px] tabular-nums transition-colors @min-[37.6rem]:inline"
```
a-frazier: **Medium** — the generation-activity toggle becomes unreachable between `22rem` and `37.6rem` of card width.

Re-gating the header on the card's container width gave the inline control and its fallback different thresholds:

- this duration chip, which toggles the activity timeline: `hidden … @min-[37.6rem]:inline` (`StoryEntry.svelte:1394`)
- the "More actions" overflow menu carrying `Show/Hide generation activity`: `@min-[22rem]:hidden` (trigger `StoryEntry.svelte:1646`, item `StoryEntry.svelte:1688`)

For any card whose content box lands in `[22rem, 37.6rem)` — i.e. `[352px, 601.6px)` — the chip has not appeared yet and the overflow menu has already gone. There is no third path: `activity.setReportVisible` is only called from those two places in this component.

Concrete case: a 1200px window with the nav panel and sidebar open narrows the story column to roughly 500–600px. A finished turn's timeline can no longer be opened, and one auto-shown during generation (`isReportVisible(id, !endedAt)`, `StoryEntry.svelte:135`) can no longer be closed. Before this change both gates were `sm:`, so exactly one was always present.

The same mismatch applies in miniature to the token badge (`@min-[23.7rem]`, `StoryEntry.svelte:1378`): on a `user_action` entry `showInfo` is false (`StoryEntry.svelte:123`), so there is no Response-info popover and the token count is invisible across `[22rem, 23.7rem)`.

The comment just above at line 1388 ("for the same reason as the token badge above") and the PR description ("then the token and duration chips hide") both describe a graded behaviour the breakpoints don't implement. Pulling the overflow fallback and the chips onto one threshold would close the gap.

src/lib/components/memory/ChapterCard.svelte
```diff
@@ -226,7 +230,11 @@
         {:else}
           <ChevronRight class="h-3.5 w-3.5" />
         {/if}
-        <span>{entries.length} entries</span>
+        <span>
+          {entries.length} entries{#if entryRange}<span class="text-muted-foreground/70 ml-1"
```
a-frazier: **Low** — a one-entry chapter renders `1 entries (#5–#5)`.

`entryRange` is emitted whenever `entries.length > 0`, so a chapter covering a single entry prints a range whose endpoints are identical, next to an already-ungrammatical count. Manual chapters can be one entry long (`createManualChapter`).

Suppressing the parenthetical when `first === last` costs one condition and removes the duplicated number.

a-frazier (commented):
Two findings from a pass over the three commits. I skipped points already raised in the existing comments (the Response-info redundancy at `StoryEntry.svelte:1731` and the `goTo` suggestion at `StoryNavPanel.svelte:79`), and left the ten-row cap removal alone.

Both are inline below.

src/lib/components/story/StoryEntry.svelte
collegatore: First, I've found that while browsing the old story chapters, I can easily get lost on where and when this happens, and opening the Info each time gets pretty annoying. Second, the Jump-to-Entry functionality often leaves me confused, even though I know precisely how it works, so I want a visual marker for "I landed where I wanted". Third, I'm working on time tracking improvements, and a part of the problem is that the time drift (which is absolutely real) often remains completely unobserved due to the fact that story time is either in the side panel or in the Info, both of which require a separate action and intention to look at. These are fairly niche cases, which is why the default for settings is "off".

collegatore:
I'm not happy with how it looks on mobile right now, so it's postponed till I find a better display

src/lib/components/memory/ChapterCard.svelte
collegatore: Fixed in 5b7c8c65: a one-entry chapter now reads `1 entry (#5)` — the duplicated endpoint is gone but the number stays, since the card exists to say which entries a chapter covers — and the count is singular.

src/lib/components/story/StoryEntry.svelte
collegatore: Fixed the Medium in 5b7c8c65: the Response info popover — the one narration control present across [22rem, 37.6rem) — now carries the activity toggle, so the report can be opened and closed at every width. The token badge on user actions is left open for the author: it has never had a fallback at any narrow width, and closing the band means reverting a threshold they tuned deliberately.

src/lib/components/layout/StoryNavPanel.svelte
collegatore: Claude backed this decision with the two identical call sites (lines 92 and 125). Can be inlined if you think that would be cleaner.

src/lib/components/story/StoryEntry.svelte
a-frazier: Fair enough!

src/lib/components/layout/StoryNavPanel.svelte
a-frazier: I'd inline this one since it's all in the script tag. The helper in ChapterEntryList keeps the markup cleaner, so it makes more sense there, even though there's only a single caller.

collegatore: Inlined in 485f4ca9: both callers now call `jumpToEntry` directly and `goTo` is gone; the helper in ChapterEntryList stays, for the markup reason you gave.

collegatore:
@a-frazier now it's ready in my opinion

src/lib/components/story/StoryEntry.svelte
```diff
@@ -135,6 +136,24 @@
...
+  let cardRect = $state<DOMRectReadOnly>()
+  const activityChipFits = $derived.by(() => {
+    if (!cardRect) return true
+    const rem = parseFloat(getComputedStyle(document.documentElement).fontSize) || 16
+    return cardRect.width >= ACTIVITY_CHIP_MIN_REM * rem
+  })
+
+  // Narration only: an action is an instant and a system entry is not story, so neither has a
+  // span to show. `retry` is narration in older saves.
+  const showEntryMeta = $derived(
```
a-frazier: `showEntryMeta` now treats `retry`-type entries as narration-like, but `showInfo` (line 123) and `activityRecord` (line 127) still gate on `entry.type === 'narration'` only. A legacy `retry` entry will show the header entry-number/time chip but has no way to open Response info or see generation activity — an inconsistent extension of the three narration-like checks.

src/lib/components/story/StoryEntry.svelte
```diff
@@ -1350,7 +1370,7 @@
         <span class="text-xs font-medium tracking-wider uppercase">System</span>
       </div>
     {:else}
-      <Icon class="text-muted-foreground h-4 w-4 shrink-0 translate-y-px" />
+      <Icon class="text-muted-foreground h-4 w-4 shrink-0" />
```
a-frazier, line 1385: The `translate-y-px` alignment class was dropped from the header icon in this (non-system, non-user_action) branch, while the sibling system-entry icon just above keeps it. Both sit in the same `flex items-center gap-2` row, so narration/retry headers now render their leading icon at a visibly different vertical offset than system-entry headers.

src/lib/components/story/StoryEntry.svelte
```diff
@@ -135,6 +136,24 @@
...
   )
 
+  // The duration chip and its fallback in Response info must never both be absent. The fallback
+  // renders in a portal outside the card, where a container query cannot reach, so one measured
+  // width decides both.
+  const ACTIVITY_CHIP_MIN_REM = 29
+  let cardRect = $state<DOMRectReadOnly>()
+  const activityChipFits = $derived.by(() => {
+    if (!cardRect) return true
+    const rem = parseFloat(getComputedStyle(document.documentElement).fontSize) || 16
```
a-frazier: `getComputedStyle(document.documentElement).fontSize` is re-parsed inside `activityChipFits` on every `cardRect` change, i.e. on every ResizeObserver tick from `bind:contentRect` — even though root font-size is effectively static per render. During a window/panel drag-resize this forces a synchronous style recalc on every tick for a value that doesn't change, adding avoidable overhead to a hot resize path.

src/lib/components/story/StoryEntry.svelte
collegatore: Fixed in b94cca2a: the three checks now share an `isNarrationLike` predicate, so a legacy `retry` entry gets Response info and generation activity alongside the header number and time.

src/lib/components/story/StoryEntry.svelte
collegatore: Fixed in b94cca2a: dropped `translate-y-px` from the system icon too, so both leading icons share one baseline in that row.

src/lib/components/story/StoryEntry.svelte
collegatore: Fixed in b94cca2a: the root font size is read once and cached in a module-level helper, so a resize tick no longer forces a style recalc.

src/lib/components/story/StoryEntry.svelte
a-frazier: This looks like a pre-existing thing.  I almost wonder if this remapping could be handled when loading entries or something, instead of having to worry about catching legacy entries everywhere.  (And that could be a separate PR.)

PR #521, opened 2026-09-18
Shrink the page for the Android keyboard, and read edge swipes off the shell

collegatore:
## What

Two platform-level fixes on Android that every screen inherits.

**The soft keyboard shrinks the page instead of covering it.** `enableEdgeToEdge()` opts the window out of the resizing `adjustResize` would do from API 30, and the soft-input mode resolved to pan. Nothing shortened the WebView, and Chromium was never told the keyboard existed: measured over DevTools on an API 36 emulator, `innerHeight` and `visualViewport.height` both stayed at the full screen with the keyboard up. The activity now takes the IME inset itself and sets it as the WebView's bottom margin, which shortens the laid-out view and with it the layout viewport. `AndroidBridge.getInsets()` reports the bottom inset net of the keyboard, so the navigation-bar strip the page reserved above it goes away. Below API 30 the window still resizes on its own, and the soft-input mode is set to resize explicitly so that path cannot resolve to pan. The bottom drawer's own `repositionInputs` is switched off, so the keyboard has one owner.

**Edge swipes are read off the shell.** The invisible strips that caught edge swipes took every tap that landed on them, including the outer half of the header's panel button and of Send on a phone. The shell's own swipe action now receives every gesture and the handlers keep only those that start within the edge band.

## Why a reviewer should look at

- `MainActivity.kt`: the inset listener passes the insets on to the view (`ViewCompat.onApplyWindowInsets`) so Chromium keeps seeing the cutout and system bars.
- `docs/architecture/overview.md`, "Edge swipes" and "The soft keyboard": the rationale, including the alternatives that were tried and failed (padding the WebView, `interactive-widget=resizes-content`).
- A panel must never reach the opposite edge's band, since swipes inside a panel bubble up to the shell. On a phone neither does: the sidebar is capped at 288px, the navigation panel at `100vw - 3rem`.

## Verified

Emulator, API 36, story screen with the action input focused:

| State | innerHeight | visualViewport | `--sab` |
| --- | --- | --- | --- |
| before, keyboard up | 800 | 800 | 48px |
| after, keyboard up | 497 | 497 | 0px |
| after, keyboard closed | 800 | 800 | 48px |

`npm run check`, `npm run lint` and `npm test` pass. Not yet verified on a real phone, nor on a device below Android 11.

## Summary by CodeRabbit

- **New Features**
  - Added edge-swipe gestures for opening the sidebar or navigation panel while viewing a story.
  - Swipe detection adapts to desktop and mobile edge widths and only opens panels when appropriate.

- **Bug Fixes**
  - Improved drawer input handling by disabling automatic input repositioning on Android while retaining it on other platforms.
  - Prevented swipe gestures from triggering when they start in text-entry fields.


src-tauri/gen/android/app/src/main/java/com/karelian/aventura/MainActivity.kt
```diff
@@ -31,6 +33,9 @@ class MainActivity : TauriActivity() {
   override fun onCreate(savedInstanceState: Bundle?) {
     enableEdgeToEdge()
     super.onCreate(savedInstanceState)
+    // Honoured below API 30, where the window still resizes for the keyboard. From API 30
+    // edge-to-edge ignores it and the margin set in onWebViewCreate does the same job.
+    window.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE)
```
a-frazier, line 38: We can probably bump the minimum API version to 30 for Android if it simplifies the code. I don't know that we need to worry about Android 8-10 device compatibility anymore.

src/lib/components/ui/drawer/drawer.svelte
```diff
@@ -7,6 +7,8 @@
     activeSnapPoint = $bindable(null),
     handleOnly = true,
     closeThreshold = 0.75,
+    // The keyboard shrinks the layout viewport itself; see overview.md, "The soft keyboard".
+    repositionInputs = false,
```
a-frazier, line 11: `repositionInputs` is defaulted to `false` for every consumer of this drawer, but the justification ("the keyboard shrinks the layout viewport itself") is Android-specific — it depends on the WebView margin trick in `MainActivity.kt`. This wrapper is shared by non-Android call sites too (SettingsModal, SyncModal, DebugLogModal, STChatImportModal, UpdateDialog, LorebookDebugPanel, PromptPackEditor), none of which gate on the existing `isAndroid()` helper (`src/lib/utils/platform.ts`). No effect today since there's no soft keyboard on desktop, but this silently disables vaul's own keyboard-avoidance globally rather than scoping it to Android — worth a look if a future iOS build or another environment reintroduces the `innerHeight`/`visualViewport` divergence.

src/lib/components/layout/AppShell.svelte
```diff
@@ -35,12 +35,24 @@
 
   const log = createLogger('AppShell')
 
-  function handleEdgeSwipeLeft() {
+  const edgeWidth = () => (window.innerWidth <= 768 ? 30 : 20)
```
a-frazier, line 38: `edgeWidth()` hardcodes the mobile breakpoint as `768` rather than reusing `DESKTOP_BREAKPOINT` from `src/lib/constants/layout.ts`, which `Sidebar.svelte` already imports for the identical desktop/mobile check. This diff already imports other constants from that module, so nothing blocks pulling this one in too. As written, the breakpoint now has three independent copies (this literal, the CSS `@media (max-width: 768px)` below, and `DESKTOP_BREAKPOINT`) — if it's ever tuned, it's easy to update the constant and miss this one, silently mismatching the edge-swipe band from the actual layout switch.

a-frazier (commented):
Two additional findings from a background review, checked against the existing thread to avoid overlap.

src/lib/components/ui/drawer/drawer.svelte
collegatore: Scoped to Android in 0675f2e9: the default is now `!isAndroid()`, so vaul keeps its own keyboard handling everywhere else. The overview doc says the same.

src/lib/components/layout/AppShell.svelte
collegatore: Switched to `DESKTOP_BREAKPOINT` in 0675f2e9. It is compared with `<=` like `ui.svelte.ts`, so the band still matches the CSS media query.

src-tauri/gen/android/app/src/main/java/com/karelian/aventura/MainActivity.kt
collegatore: That's a one-liner, so raising the minimal API will probably have the same amount of changes but with bigger consequences, and I'm not fond of deciding it myself.

